PR #4, opened 2023-10-26
html-css/Niloufar

Niloufar97:


1_html-css/index.html
```diff
@@ -0,0 +1,214 @@
...
+        <div class="logo">Niloufar Khaefi</div>
+        <div>
+            <ul class="nav-links">
+                <li><a href="#about">About</a></li>
+                <li><a href="#skills">Skills</a></li>
+                <li><a href="#projects">Projects</a></li>
+                <li><a href="#contact">Contact</a></li>
+            </ul>
+        </div>
+    </nav>
```
sergii-borodin: In that particular structure, to have the nav tag outside of the header tag is the wrong decision. An example of having the nav tag outside could be a navigation on a sidebar.
I would strongly suggest you put the nav tag inside the header.

1_html-css/index.html
```diff
@@ -0,0 +1,214 @@
...
+            <div class="profile-btn">
+                <button class="btn cv-btn" onclick="window.open('./assets/Niloufar-cv.pdf')">Download CV</button>
+                <button class="btn contact-btn" onclick="location.href = '#contact'">Contact Info</button>
+            </div>
+            <div class="SoMe-links">
+                <a href="https://github.com/Niloufar97"><img src="assets/square-github.svg" alt=""></a>
+                <a href="https://www.linkedin.com/in/niloufar-khaefi/"><img src="assets/linkedin.svg" alt=""></a>
+            </div>
+        </div>
+    </header>
```
sergii-borodin: Instead of the header tag here, I would recommend using the section. A common practice would be to call this section 'Hero'. Try to google to get information about the hero section😊.

1_html-css/index.html
```diff
@@ -0,0 +1,214 @@
...
+                        <h5>Experience</h5>
+                        <p>1+ year</p>
+                        <p>Frontend Development</p>
+                    </div>
+                    <div class="education">
+                        <h5>Education</h5>
+                        <p>B.Sc. Bachlors Degree</p>
+                        <p>Isfahan University of Technology </p>
+                    </div>
+                </div>
```
sergii-borodin: Instead of having a div as a container 'experience-box', use the semantic tag ul, because you want to have two entities with the same type of content inside.
Two dives  'experience' and 'education', I would change to the li tag accordingly.

1_html-css/index.html
```diff
@@ -0,0 +1,214 @@
...
+        </div>
+        <div class="skill-bar-container">
+            <div class="skill">
+                <div class="skill-name">React</div>
+                <div class="skill-bar">
+                    <div class="skill-level react"></div>
+                    <div>50%</div>
+                </div>
+            </div>
+        </div>
```
sergii-borodin: The same thing. Create a container using "ul" and put every skill in "li". In this case, you will reduce the amount of dives😄. 
Bad practice to write text within the "div" tag. The "div" tag is a container element, which allows you to group other elements together and apply styles to them. However, you should use the "p" tag for paragraphs and other long blocks of text, as it provides better semantic meaning and styling options.

1_html-css/index.html
```diff
@@ -0,0 +1,214 @@
...
+            <div class="SoMe-links">
+                <a href="https://github.com/Niloufar97"><img src="assets/square-github.svg" alt=""></a>
+                <a href="https://www.linkedin.com/in/niloufar-khaefi/"><img src="assets/linkedin.svg" alt=""></a>
+            </div>
+        </div>
+    </header>
+    <section id="about" data-aos="fade-up" data-aos-duration="2000">
+        <div class="about-title">
+            <p>Get To Know More</p>
+            <h2 class="big-font">About Me</h2>
```
sergii-borodin: If you propose to use the "p" here to style the text differently? I would recommend you to use "span".
The "span" tag is used for short blocks of text or for adding formatting to the text.

1_html-css/index.html
```diff
@@ -0,0 +1,214 @@
...
+                    <img src="assets/todo-list.png" alt="">
+                </div>
+                <div class="project-txt">
+                    <h3>Todo List</h3>
+                    <p>JavaScript</p>
+                </div>
+                <div class="project-link">
+                    <a href="https://github.com/Niloufar97/to-do-list">See Project</a>
+                </div>
+            </div>
```
sergii-borodin: The same problem("ul" and 3 "li").

1_html-css/index.html
```diff
@@ -0,0 +1,214 @@
...
+        <div class="contact-me">
+            <div class="email">
+                <img src="assets/103176_mail_email_icon.svg" alt="">
+                <a href="mailto:nilooooofar97@gmail.com">nilooooofar97@gmail.com</a>
+            </div>
+            <div class="linkdin">
+                <img src="assets/linkedin.svg" alt="">
+                <a href="https://www.linkedin.com/in/niloufar-khaefi">Niloufar-Khaefi</a>
+            </div>
+        </div>
```
sergii-borodin: Also here🤓

sergii-borodin (commented):
Hello Niloufar.
You have done an awesome job💪. 
I like it very much. 
But some things you should change to make it perfect😀.
Looking forward to seeing your changes.
Have a good one.

PR #13, opened 2023-10-27
html-css/anastasiia2

humenyuk16:


projects_portfolio/index.html
```diff
@@ -0,0 +1,172 @@
...
+    </header>
+    <section id="welcome-section" class="special-section">
+      <div class="text">
+        <h1>Hi, my name is Anastasiia</h1>
+        <p>
+          Welcome to my portfolio! I'm a passionate and driven<br />
+          junior front-end developer with a flair for creating captivating and
+          user-friendly web experiences.
+        </p>
+        <div class="botton">
```
SvitlanaBurlaka: this div can be named differently, e.g. 'button-container'

projects_portfolio/styles.css
```diff
@@ -0,0 +1,522 @@
...
+  --white: #fff;
+  --blue: #c1eafc;
+  --dark_blue: #77cbef;
+}
+@font-face {
+  font-family: "Sniglet";
+  src: url(Sniglet/Sniglet-Regular.ttf);
+}
+
+body {
```
SvitlanaBurlaka, line 15: When you inspect the page in netlify, your `body` tag has 8 pixels. To remove it you can add : margin: 0;
Screenshot 2023-10-31 at 14 57 01

projects_portfolio/index.html
```diff
@@ -0,0 +1,172 @@
+<!DOCTYPE html>
```
SvitlanaBurlaka, line 1: Overall good, I would recommend to play with responsiveness, as I can see, your idea was to create for three screen sizes, for example when width is 590px header starts behaving strange. Good work! ❤️

SvitlanaBurlaka: Also about header, I would recommend to change it to have in this order as it it on the page : Home, About me, Projects and then Contacts

PR #15, opened 2023-10-28
html-css/Sanaz

Sanazre:
This is my first HTML and CSS project,  Thank you.


styles.css
```diff
@@ -1,539 +1,179 @@
...
-/*Section 1
-welcome-section */
 
-#welcome-section {
-  display: flex;
-  flex: 1;
-  height: 100vh;
-  justify-content: center;
-  align-items: center;
+#welcome-text {
```
SvitlanaBurlaka, line 40: I would recommend to change color or move this p tag somewhere else, so it has better visibility 😃

index.html
```diff
@@ -3,175 +3,101 @@
...
-    <section id="projects">
-      <div class="projects-header">
-      <h2>My Projects</h2>
-      </div>
-      <div class="projects-link">
-      <div class="survey_form_project">
+    <section id="skills">
+      <h3>Skills</h3>
+
+      <li>HTML5</li>
```
SvitlanaBurlaka, line 51: `li` elements must be wrapped in `ul` tag

SvitlanaBurlaka (commented):
I will recommend to add icons e.g. for social media, also to style list with your skills [https://sharkcoder.com/blocks/list](here maybe you can find inspiration)

PR #43, opened 2023-11-01
git-week1/Helena

helena69hme:


sergii-borodin (commented):
Hi Helena😊.
Good job💪.
One nitpick. The messages for commits could be more descriptive. It helps you and your team understand the purpose and context of a particular change in your codebase.

Instead of 'favourite number1', you could write 'add my first favourite food recipe'.
The same way works for the second commit message.
And for the third commit message, could be 'add a list of visited countries'.

PR #47, opened 2023-11-02
html-css/Monika

Monika-Marciniak:


1_html-css/contact.html
```diff
@@ -0,0 +1,35 @@
...
+      <div>
+        <img src="./linkedin-icon.png" />
+      </div>
+    </div>
+    <div class="name"></div>
+    <section>
+      <div>
+        <h3>You can contact me via:</h3>
+        <h2>email m.marciniak89@gmail.com</h2>
+        <h2>https://www.linkedin.com/in/monika-marciniak2/</h2>
```
julczka, line 28: If you want this elements to link somewhere and be clickable, you need to wrap the address in an anchor tag (<a></a>). It also might not be a best idea to put links inside the headings tags: due to accessibility it is probably best to leave the h tags for headings and just style the anchor tags how you want them to look. The structure of your html determines how the webpage is read by screen readers thats people with poor eyesight may use to read the text, and for the screen reader it is important that elements are used correctly. This is of course not that important with exercises but it's good to have that in mind in the future, to make your work accessible for everyone.

1_html-css/skills.html
```diff
@@ -0,0 +1,39 @@
...
+          </ul>
+        </div>
+        <div>
+          <img src="./linkedin-icon.png" />
+        </div>
+      </div>
+      <div class="name"></div>
+
+      <div class="begginer">
+        <h1>HTML begginer</h1>
```
julczka, line 27: According to accessibility rules, you should have only one h1 per page. That is also better for optimizing a page for SEO - it will end up higher in Google search if you stick to a set of rules. All pages are eventually read by search engines crawlers - automatic readers that go through the code and index what's in it. It's important to follow the SEO rules if you're making a web shop for example - you want it to end up high in search results. That is again not that important for the exercise purpose but good to keep in mind that something like that exists.

julczka (approved):

PR #69, opened 2023-11-09
Javascript javascript1 week1/gizem

GizemSavci:


3_javascript/javascript1/week1/main.js
```diff
@@ -38,4 +38,15 @@ function  calculator(wide, deep, high, gardenSizeInM2, price){
...
+
+//Ez Namey
+const firstWords = ["Cutie", "Tabby", "Chonky", "Grumpy", "Liquid", "Fluffy", "Squishable", "One-Brain-Cell", "Hangry", "Sleepy"]
+const secondWords = ["Cat", "Goat", "Giraffe", "Lion", "Tiger", "Cow", "Horse", "Duck", "Whale", "Leopard"]
+const randomNumber = Math.floor(Math.random() * 10);
+
+const startupName = firstWords[randomNumber] + " " + secondWords[randomNumber]
+console.log(startupName)
+
+console.log("The startup: " + startupName + " contains " + startupName.length + " characters.")
```
SvitlanaBurlaka, line 52: your `startupName.length` will give you always number with an space between two words, wich you adding here : `const startupName = firstWords[randomNumber] + " " + secondWords[randomNumber] `.
So to get precise number of characters without space, you can use for example : `console.log("The startup: " + startupName + " contains " + (startupName.length-1) + " characters.")` :)

3_javascript/javascript1/week1/main.js
```diff
@@ -4,3 +4,18 @@ let yearFuture = 2045;
...
+
+
+//Goodboy-Oldboy
+let dogYearOfBirth = 2023;
+let dogYearFuture = 2027;
+let dogYear = 7;
+let shouldShowResultInDogYears = true;
+let dogAge = dogYearFuture - dogYearOfBirth;
+
+if (shouldShowResultInDogYears === true){
```
SvitlanaBurlaka, line 16: you do not need to declare in an if statement that `shouldShowResultInDogYears === true`, simply will be enough to write :`if (shouldShowResultInDogYears ){...` because you declared it above .

PR #71, opened 2023-11-11
javascript1-week1/Viki

VictoriaDem17:
Finished my homework for javascript1-week1.

### DIFF
--- a/3_javascript/javascript1/week1/smart-ease.js
+++ b/3_javascript/javascript1/week1/smart-ease.js
@@ -1,0 +1,26 @@
+console.log("Javascript/week1");
+//Age-ify (A future age calculator)
+const yearOfBirth = 1992;
+let yearFuture = 2024;
+const age = yearFuture - yearOfBirth;
+console.log(age);
+console.log("You will be " + age + " years old in " + yearFuture);
+
+//Goodboy-Oldboy (A dog age calculator)
+const dogYearOfBirth = 2017;
+let dogYearFuture = 2027;
+let dogYear = dogYearFuture - dogYearOfBirth;
+console.log(dogYear);
+let dogYearsCompareToHuman = dogYear * 7;
+console.log(dogYearsCompareToHuman);
+shouldShowResultInDogYears = true;
+shouldShowResultInDogYears = false;
+function test(shouldShowResultInDogYears) {
+    if (shouldShowResultInDogYears) {
+        return console.log("Your dog will be " + dogYear + " human years old in " + dogYearFuture);
+    }
+        return console.log("Your dog will be " + dogYearsCompareToHuman + " dog years old in " + dogYearFuture); 
+}
+
+test(true);
+test(false);
--- a/3_javascript/javascript1/week1/smart-ease.js
+++ b/3_javascript/javascript1/week1/smart-ease.js
@@ -24,3 +24,59 @@ function test(shouldShowResultInDogYears) {
 
 test(true);
 test(false);
+
+//Housey pricey (A house price estimator)
+
+//House for Peter
+const housePriceForPeter = 2500000;
+let houseWideForPeter = 8;
+let houseDeepForPeter = 10;
+let houseHighForPeter =10;
+let houseGardenSizeForPeter = 100;
+const volumeInMetersForPeter = 8*10*10*2.5*1000;
+console.log(volumeInMetersForPeter);
+const realHousePriceForPeter = volumeInMetersForPeter+(houseGardenSizeForPeter*300);
+console.log(realHousePriceForPeter);
+console.log("The real price for Peter during calculation is " + realHousePriceForPeter + ", and it means that he is paying much more " + "(" + housePriceForPeter + ")" + " for the house with those parameters.");
+
+//House for Julia
+const housePriceForJulia = 1000000;
+let houseWideForJulia = 5;
+let houseDeepForJulia = 11;
+let houseHighForJulia = 8;
+let houseGardenSizeForJulia = 70;
+const volumeInMetersForJulia = 5*11*8*2.5*1000;
+console.log(volumeInMetersForJulia);
+const realHousePriceForJulia = volumeInMetersForJulia+(houseGardenSizeForJulia*300);
+console.log(realHousePriceForJulia);
+console.log("The real price for Julia during calculation is " + realHousePriceForJulia + ", and it means that she is paying much less " + "(" + housePriceForJulia + ")" + " for the house with those parameters.");
+
+
+//Ez Namey (Startup name generator)
+let firstWords = [
+    "Funny",
+    "Ambitious",
+    "Intelligent",
+    "Funny",
+    "Warm-hearted",
+    "Cheerful",
+    "Unique",
+    "Charming",
+    "Attractive",
+    "Pro-active"
+];
+let secondWords = [
+    "Donkey",
+    "Wagon",
+    "Cow",
+    "Window",
+    "Badger",
+    "Space",
+    "Genius",
+    "Hedgehog",
+    "Plant",
+    "Bees"
+];
+const randomNumber = Math.floor(Math.random() * 10);
+const startupName = firstWords[randomNumber] + " " + secondWords[randomNumber];
+console.log("The startup: " + startupName + " contains " + startupName.length + " characters");
--- a/3_javascript/javascript1/week1/smart-ease.js
+++ b/3_javascript/javascript1/week1/smart-ease.js
@@ -1,29 +1,27 @@
 console.log("Javascript/week1");
 //Age-ify (A future age calculator)
 const yearOfBirth = 1992;
-let yearFuture = 2024;
+const yearFuture = 2024;
 const age = yearFuture - yearOfBirth;
 console.log(age);
 console.log("You will be " + age + " years old in " + yearFuture);
 
 //Goodboy-Oldboy (A dog age calculator)
 const dogYearOfBirth = 2017;
-let dogYearFuture = 2027;
-let dogYear = dogYearFuture - dogYearOfBirth;
+const dogYearFuture = 2027;
+const dogYear = dogYearFuture - dogYearOfBirth;
 console.log(dogYear);
-let dogYearsCompareToHuman = dogYear * 7;
+const dogYearsCompareToHuman = dogYear * 7;
 console.log(dogYearsCompareToHuman);
-shouldShowResultInDogYears = true;
-shouldShowResultInDogYears = false;
-function test(shouldShowResultInDogYears) {
+function dogsAge(shouldShowResultInDogYears) {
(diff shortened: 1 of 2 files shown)


3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
+console.log("Javascript/week1");
+//Age-ify (A future age calculator)
+const yearOfBirth = 1992;
+let yearFuture = 2024;
```
BrantalikP: `let` keyword we usually use only if we want to re-assign the variable later, but in our case, it's better to use constant `const`, because we want to keep the value the same for the whole time
e.g
```
let name = "peter";

console.log(name); // peter

name = "lucas";

console.log(name); // lucas


```

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+
+//Goodboy-Oldboy (A dog age calculator)
+const dogYearOfBirth = 2017;
+let dogYearFuture = 2027;
+let dogYear = dogYearFuture - dogYearOfBirth;
+console.log(dogYear);
+let dogYearsCompareToHuman = dogYear * 7;
+console.log(dogYearsCompareToHuman);
+shouldShowResultInDogYears = true;
+shouldShowResultInDogYears = false;
```
BrantalikP:
```
shouldShowResultInDogYears = true;
shouldShowResultInDogYears = false;
```

To make it work properly, the variable must be declared first as `let shouldShowResultInDogYears;`, but since you are not using those variables at all, you can delete them ;)

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+const yearOfBirth = 1992;
+let yearFuture = 2024;
+const age = yearFuture - yearOfBirth;
+console.log(age);
+console.log("You will be " + age + " years old in " + yearFuture);
+
+//Goodboy-Oldboy (A dog age calculator)
+const dogYearOfBirth = 2017;
+let dogYearFuture = 2027;
+let dogYear = dogYearFuture - dogYearOfBirth;
```
BrantalikP: the logic is correct!👏 Just replace the `let` with `const` keywords as I suggested above, please :)

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+shouldShowResultInDogYears = false;
+function test(shouldShowResultInDogYears) {
+    if (shouldShowResultInDogYears) {
+        return console.log("Your dog will be " + dogYear + " human years old in " + dogYearFuture);
+    }
+        return console.log("Your dog will be " + dogYearsCompareToHuman + " dog years old in " + dogYearFuture); 
+}
+
+test(true);
+test(false);
```
BrantalikP: The logic itself is correct! 💪 but naming a function just with plain `test` can cause some confusion about its functionality if the code gets bigger and bigger, therefore I suggest renaming it to something more specific e.g `testDogYer`

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+    "Badger",
+    "Space",
+    "Genius",
+    "Hedgehog",
+    "Plant",
+    "Bees"
+];
+const randomNumber = Math.floor(Math.random() * 10);
+const startupName = firstWords[randomNumber] + " " + secondWords[randomNumber];
+console.log("The startup: " + startupName + " contains " + startupName.length + " characters");
```
BrantalikP: Excellent! Good job! The only caveat here is that when you are counting the characters in ﻿`startupName` with `﻿startupName.length`, it also includes the space (﻿+ " " +). Since we do not want to include this in our value, we have two options:

1.	Declare the names separately and then add them together.
```
const firstWord = firstWords[randomNumber] 
const secondWord = secondWords[randomNumber];
const numOfCharacters = firstWord.length + secondWord.length
console.log("The startup: " + startupName + " contains " + numOfCharacters + " characters");
```

2. simple substruct 1 from the value like this 

`console.log("The startup: " + startupName + " contains " + startupName.length - 1 + " characters");`

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+let houseGardenSizeForJulia = 70;
+const volumeInMetersForJulia = 5*11*8*2.5*1000;
+console.log(volumeInMetersForJulia);
+const realHousePriceForJulia = volumeInMetersForJulia+(houseGardenSizeForJulia*300);
+console.log(realHousePriceForJulia);
+console.log("The real price for Julia during calculation is " + realHousePriceForJulia + ", and it means that she is paying much less " + "(" + housePriceForJulia + ")" + " for the house with those parameters.");
+
+
+//Ez Namey (Startup name generator)
+let firstWords = [
```
BrantalikP: `let` keyword replace with `const`

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+    "Intelligent",
+    "Funny",
+    "Warm-hearted",
+    "Cheerful",
+    "Unique",
+    "Charming",
+    "Attractive",
+    "Pro-active"
+];
+let secondWords = [
```
BrantalikP: `let` keyword replace with `const`

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+
+//Housey pricey (A house price estimator)
+
+//House for Peter
+const housePriceForPeter = 2500000;
+let houseWideForPeter = 8;
+let houseDeepForPeter = 10;
+let houseHighForPeter =10;
+let houseGardenSizeForPeter = 100;
+const volumeInMetersForPeter = 8*10*10*2.5*1000;
```
BrantalikP: Hey I see that you have declared the house values, but you are not actually using it :) 

`const volumeInMetersForPeter = 8*10*10*2.5*1000;`

instead of this it should count also just the volumeInMeters so `*2.5*1000;` should not be part of this variable

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+console.log(volumeInMetersForPeter);
+const realHousePriceForPeter = volumeInMetersForPeter+(houseGardenSizeForPeter*300);
```
BrantalikP: the equation is `housePrice = volumeInMeters * 2.5 * 1000 + gardenSizeInM2 * 300;` try to recreate this structure with your variables for `realHousePriceForPeter`

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+console.log(realHousePriceForPeter);
+console.log("The real price for Peter during calculation is " + realHousePriceForPeter + ", and it means that he is paying much more " + "(" + housePriceForPeter + ")" + " for the house with those parameters.");
```
BrantalikP: to check wheter the house is too expensive or not you should use do the `if else` statement (same as with `dogYears`)  so something like 

```
if(housePriceForPeter > realHousePriceForPeter) {
// too expensive
} else {
// good deal
}
```

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,82 @@
...
+const housePriceForJulia = 1000000;
+let houseWideForJulia = 5;
+let houseDeepForJulia = 11;
+let houseHighForJulia = 8;
+let houseGardenSizeForJulia = 70;
+const volumeInMetersForJulia = 5*11*8*2.5*1000;
+console.log(volumeInMetersForJulia);
+const realHousePriceForJulia = volumeInMetersForJulia+(houseGardenSizeForJulia*300);
+console.log(realHousePriceForJulia);
+console.log("The real price for Julia during calculation is " + realHousePriceForJulia + ", and it means that she is paying much less " + "(" + housePriceForJulia + ")" + " for the house with those parameters.");
```
BrantalikP: The same comments as with Petrs calculations :)

BrantalikP (changes requested):
Amazing job @VictoriaDem17! 👏 I love that you separated each logic with a comment, this is really important to keep the code clean and readable, especially when you are working on a team. I added some comments, but please, do not be afraid by the number of them, those are just small things and if you have some questions, feel free to reach out to me. Petr Brantalik on Slack 💪

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,87 @@
...
+dogsAge(false);
+
+//Housey pricey (A house price estimator)
+
+//House for Peter
+const housePriceForPeter = 2500000;
+let houseWideForPeter = 8;
+let houseDeepForPeter = 10;
+let houseHighForPeter =10;
+let houseGardenSizeForPeter = 100;
```
BrantalikP, line 33: these are remain constant as well, so you can replace `let` here with `const` also :)

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,87 @@
...
+    else {
+    console.log("Good deal.")
+    }
+
+//House for Julia
+const housePriceForJulia = 1000000;
+let houseWideForJulia = 5;
+let houseDeepForJulia = 11;
+let houseHighForJulia = 8;
+let houseGardenSizeForJulia = 70;
```
BrantalikP, line 49: Here as well :)

3_javascript/javascript1/week1/smart-ease.js
```diff
@@ -0,0 +1,87 @@
...
+    "Badger",
+    "Space",
+    "Genius",
+    "Hedgehog",
+    "Plant",
+    "Bees"
+];
+const randomNumber = Math.floor(Math.random() * 10);
+const startupName = firstWord[randomNumber] + " " + secondWord[randomNumber];
+console.log("The startup: " + startupName + " contains " + (startupName.length-1) + " characters");
```
BrantalikP, line 87: great job! This is not an issue, but maybe next time for better readablity you can extract `startupName.length-1` to a separate variable as 
```
const lettersCount = startupName.length-1 
console.log("The startup: " + startupName + " contains " + lettersCount + " characters");
```

BrantalikP (approved):
Amazing job, @VictoriaDem17! You managed to resolve all comments and take this homework to the next level!
I really appreciate that you came back to finish this. It's really important, not just in the IT world, but in any field, to see tasks through to the end. Kudos for that 👏.

PR #72, opened 2023-11-11
javascript1-week1/Helena

helena69hme:
Javascript tasks done!

### DIFF
--- a/3_javascript/javascript1/week1/javascript.js
+++ b/3_javascript/javascript1/week1/javascript.js
@@ -1,0 +1,104 @@
+ //hello... Hello... This is my first experience in Javascript :)
+
+console.log("******    Age-ify (A future age calculator)    ******");
+let myYearOfBirth = 1991;
+let yearFuture = 2027;
+let age = yearFuture - myYearOfBirth;
+
+console.log("I\'ll be " + age + " years old in " + yearFuture + ".");
+console.log("\n");
+
+
+
+console.log("******    Goodboy-Oldboy (A dog age calculator)    ******");
+let dogYearOfBirth = 2017;
+let dogYearFuture = 2045;
+// We have a conditional statement, so a "BOOLEAN variable" can represent (shouldShowResultInDogYears).
+let shouldShowResultInDogYears = true;
+
+ // According to the information in the txt:(1 human year = 7 dog year)
+let dogAge = (dogYearFuture - dogYearOfBirth)*7;
+let humanAge = dogYearFuture - dogYearOfBirth;
+
+if (shouldShowResultInDogYears) {
+    console.log("His dog will be " + dogAge + " dog years old in " + dogYearFuture + ".");
+}
+
+else {
+ console.log("His dog will be " + humanAge + " human years old in " + dogYearFuture + ".");
+}
+console.log("/* His dog will be 28 human years old in 2045. */");
+console.log("\n");
+
+
+console.log("******    Housey pricey (A house price estimator)    ******");
+//Peter:
+var petersPreferredHouseWide = a;
+var petersPreferredHouseDeep = b;
+var petersPreferredHouseHigh = c;
+
+let petersGardenSizeInM2 = 100; //m2
+let petersPreferredHouseCosts = 2500000; //$
+
+var a = 8; //m
+var b = 10; //m
+var c = 10; //m
+
+var volumeInMeters = a * b * c;
+const petersHousePrice = volumeInMeters * 2.5 * 1000 + petersGardenSizeInM2 * 300;
+const petersPriceDifference = petersPreferredHouseCosts - petersHousePrice;
+//We can consider the currency unit as the US Doller.
+console.log("Peter is paying " + petersPriceDifference + " more than real price.");
+
+//Julia:
+var juliasPreferredHouseWide = x;
+var juliasPreferredHouseDeep = y;
+var juliasPreferredHouseHigh = z;
+
+let juliasGardenSizeInM2 = 70; //m2
+let juliasPreferredHouseCosts = 1000000; //$
+
+var x = 5; //m
+var y = 11; //m
+var z = 8; //m
+
+var volumeInMeters = x * y * z;
+const juliasHousePrice = volumeInMeters * 2.5 * 1000 + juliasGardenSizeInM2 * 300;
+const juliasPriceDifference = juliasPreferredHouseCosts - juliasHousePrice;
+//We can consider the currency unit as the US Doller.
+console.log("Julia is paying " + juliasPriceDifference + " less than real price.");
+console.log("\n");
+
+
+console.log("******    Ez Namey (Startup name generator) Optional    ******");
+const firstWords =  [
+    "Golden",
+    "Amazing",
+    "Brilliant",
+    "Cute",
+    "Inteligent",
+    "Smart",
+    "Silver",
+    "Shine",
+    "Alpha",
+    "Fast"
+];
+
+const secondWords = [
+    "Magic",
+    "Mind",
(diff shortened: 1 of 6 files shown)


3_javascript/javascript1/week1/javascript.js
```diff
@@ -0,0 +1,104 @@
+ //hello... Hello... This is my first experience in Javascript :)
+
+console.log("******    Age-ify (A future age calculator)    ******");
+let myYearOfBirth = 1991;
+let yearFuture = 2027;
+let age = yearFuture - myYearOfBirth;
```
BrantalikP: Great Job! Just regarding using the `let` keyword.  We usually use it only if we want to re-assign the variable later, but in our case, it's better to use constant `const,` because we want to keep the value the same for the whole time
example of using `let`
```
let name = "peter";

console.log(name); // peter

name = "lucas";

console.log(name); // lucas


```

3_javascript/javascript1/week1/javascript.js
```diff
@@ -0,0 +1,104 @@
...
+let humanAge = dogYearFuture - dogYearOfBirth;
+
+if (shouldShowResultInDogYears) {
+    console.log("His dog will be " + dogAge + " dog years old in " + dogYearFuture + ".");
+}
+
+else {
+ console.log("His dog will be " + humanAge + " human years old in " + dogYearFuture + ".");
+}
+console.log("/* His dog will be 28 human years old in 2045. */");
```
BrantalikP: Great! again, just replace the `let` keywords with `const` and since you are already stating the solution within the condition, you can delete the hardcoded lines

```
console.log("/* His dog will be 28 human years old in 2045. */");
console.log("\n");
```

3_javascript/javascript1/week1/javascript.js
```diff
@@ -0,0 +1,104 @@
...
+console.log("\n");
+
+
+console.log("******    Housey pricey (A house price estimator)    ******");
+//Peter:
+var petersPreferredHouseWide = a;
```
BrantalikP: - `var`  keyword is not really used in javascript anymore, we use only `const` or `let` these days, please replace the `var` keywords with `const` 

- this is not an issue anymore but since the Javascript is synchronous language (its read from top to bottom) 
we should always first declare a value and then use it as this 
```
var a = 8; //m
var b = 10; //m
var c = 10; //m

const petersPreferredHouseWide = a;
const petersPreferredHouseDeep = b;
const petersPreferredHouseHigh = c;
```
Also, I think using a,b,c variables is a bit redundant here. you can simply assign the value directly to the variables as

```
const petersPreferredHouseWide = 8;
const petersPreferredHouseDeep = 10;
const petersPreferredHouseHigh = 10;
```

3_javascript/javascript1/week1/javascript.js
```diff
@@ -0,0 +1,104 @@
...
+let petersGardenSizeInM2 = 100; //m2
+let petersPreferredHouseCosts = 2500000; //$
+
+var a = 8; //m
+var b = 10; //m
+var c = 10; //m
+
+var volumeInMeters = a * b * c;
+const petersHousePrice = volumeInMeters * 2.5 * 1000 + petersGardenSizeInM2 * 300;
+const petersPriceDifference = petersPreferredHouseCosts - petersHousePrice;
```
BrantalikP: I think, that you should check whether the price is too high or the price is good using a condition (same as with `dogYears` assignment)

something like 
```
if(housePrice > petersHousePrice) {
 // too expensive } else {
 // good deal }
```

3_javascript/javascript1/week1/javascript.js
```diff
@@ -0,0 +1,104 @@
...
+let juliasGardenSizeInM2 = 70; //m2
+let juliasPreferredHouseCosts = 1000000; //$
+
+var x = 5; //m
+var y = 11; //m
+var z = 8; //m
+
+var volumeInMeters = x * y * z;
+const juliasHousePrice = volumeInMeters * 2.5 * 1000 + juliasGardenSizeInM2 * 300;
+const juliasPriceDifference = juliasPreferredHouseCosts - juliasHousePrice;
```
BrantalikP: same comments as with Petr calculations :)

3_javascript/javascript1/week1/javascript.js
```diff
@@ -0,0 +1,104 @@
...
+    "Maad",
+    "Journey",
+    "Team",
+    "Engeener",
+    "Code"
+];
+
+const randomNumber = Math.floor(Math.random() * 10);
+console.log(randomNumber);
+const startupName = firstWords[randomNumber] + " " + secondWords[randomNumber];
```
BrantalikP: Good Job @helena69hme! The only one caveat is counting the characters in ﻿`startupName` with `﻿startupName.length`, it also includes the space (﻿+ " " +). Since we do not want to include this in our value, we have two options:

1. Declare the names separately and then add them together.
```
const firstWord = firstWords[randomNumber] 
const secondWord = secondWords[randomNumber];
const numOfCharacters = firstWord.length + secondWord.length
console.log("The startup: " + startupName + " contains " + numOfCharacters + " characters");
```
2. simple substruct 1 from the value like this
`console.log("The startup: " + startupName + " contains " + startupName.length - 1 + " characters");`

BrantalikP (changes requested):
@helena69hme Great Job! Really like your approach, the code looks structured and clean. Please do get scared by the comments,  the most of them are just suggestions. you should be proud of yourself 👏 If you have some questions, please feel free to reach out to me!

3_javascript/javascript1/week1/javascript.js
helena69hme: great explanation... love it :)
done!

3_javascript/javascript1/week1/javascript.js
helena69hme: done!

helena69hme: just did not know should deleted this part also or not :
" console.log("\n") ;"

BrantalikP: yes, you can delete it as well :)

3_javascript/javascript1/week1/javascript.js
helena69hme: many thanks for that much attention. 
Done!

3_javascript/javascript1/week1/javascript.js
helena69hme: Done!

3_javascript/javascript1/week1/javascript.js
helena69hme: Dear @BrantalikP
I followed up your advice step by step...Done now although it was a little challenging. I'm a bit confused.
Let me ask you why by using the second substruct which you offered, this message appeares on the js console:
"NaN characteres"

js length
jv length concole

BrantalikP: Hey @helena69hme 

Sorry this is my mistake, I wrote the code here just as plain example and did not realize the mistake I made.

You should subtract the number first in separate variable, then use it inside of the string. 

like this:

```
 const count = startupName.length - 1;
 console.log("The startup: " + startupName + " contains " + count + " characters");
```

this should work :)

helena69hme: oh!  I see :)
you are not who should be sorry... I must persist in finding the answer, with more practice, deeper exploration, and trying different paths. I am infinitely grateful for your explanation and attention.

BrantalikP (approved):
Great Job! 👏

PR #73, opened 2023-11-11
javascript1-week1/maryam

MaryamAarhus:


3_javascript/javascript1/javaScript/script.js
```diff
@@ -0,0 +1,71 @@
...
+console.log("Your dog will be " + dogYear + "dog years old in " + dogYearFuture);
+
+//Price House Estimator//
+
+//Peter's house details//
+let houseWidthPeter = 8;
+let houseDepthPeter = 10;
+let houseHeightPeter = 10;
+let gardenSizeInM2Peter = 100;
+let houseCostPeter = 2.500000;
```
SvitlanaBurlaka, line 27: I would suggest to use here `const` instead of `let`, since those values is not going to change, right? 😄

3_javascript/javascript1/javaScript/script.js
```diff
@@ -0,0 +1,71 @@
...
+
+//Julia's house details//
+let houseWidthJulia = 5;
+let houseDepthJulia = 11;
+let houseHeightJulia = 8;
+let gardenSizeInM2julia = 70;
+let houseCostJulia = 1.000000;
```
SvitlanaBurlaka, line 34: here as well, as I commented above, I think that better is `const`

3_javascript/javascript1/javaScript/script.js
```diff
@@ -0,0 +1,71 @@
...
+let volumeJulia = houseWidthJulia * houseDepthJulia * houseHeightJulia;
+
+//Claculate house pricing using the following formula//
+let housePricePeter = volumePeter * 2.5 * 1000 + gardenSizeInM2Peter * 300;
+let housePriceJulia = volumeJulia * 2.5 * 1000 + gardenSizeInM2julia * 300;
+
+//Decide if Peter and Julia are paying too much/little//
+console.log("Peter's house should cost " + housePricePeter);
+console.log("Julia's house should cost " + housePriceJulia);
+console.log("Peter is paying too " + (houseCostPeter > housePricePeter ? "much" : "little."));
```
SvitlanaBurlaka, line 48: good use of ternary operators! ⭐

SvitlanaBurlaka (commented):
Good job! ⭐  I would just recommend one thing : when creating a variable, think always if the value is going to change or not, and then decide based on that to use `const` or `let`. I know, on this stage it is a bit confusing, but later on it will get more clear 😄

PR #74, opened 2023-11-11
javascript1-week1/Ihor

IhorKulytskyi:
Finished my homework

### DIFF
--- a/3_javascript/javascript1/week1/Startup-name.js
+++ b/3_javascript/javascript1/week1/Startup-name.js
@@ -3,5 +3,5 @@ const firstWords = ["Easy", "Awesome", "Corporate", "Super", "Good"];
         const randomNumber = Math.floor(Math.random() * 5);     
         let startupName = firstWords[randomNumber] + " " + secondWords[randomNumber];
         console.log(startupName);
-        let length = startupName.length
+        let length = startupName.length - 1;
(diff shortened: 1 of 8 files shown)


3_javascript/javascript1/week1/Startup-name.js
```diff
@@ -0,0 +1,7 @@
+const firstWords = ["Easy", "Awesome", "Corporate", "Super", "Good"];
+        const secondWords = ["Corporation", "City", "Working", "Power", "World"];
+        const randomNumber = Math.floor(Math.random() * 5);     
+        let startupName = firstWords[randomNumber] + " " + secondWords[randomNumber];
+        console.log(startupName);
+        let length = startupName.length
```
SvitlanaBurlaka: your length variable contains also an empty space between words, which is not actually a character , to have precise number of character you can use this, for example : `let length = startupName.length - 1`

SvitlanaBurlaka (commented):
@IhorKulytskyi  Great work, I have nothing else to add. I can see, that you used separated files for each exercise, which is also right. But in my opinion, it can be more structured if you would have all exercises in one js file, and it is easier to navigate through the homework , but of course it is up to you 😸

IhorKulytskyi:
Thanks for the feedback. I've changed the "empty space between words".

PR #82, opened 2023-11-13
javascript1-week1/Daisy

Daisyangtan:
Hi! Here's my assignment for last week. Let me know if I did something wrong. Thank you!

### DIFF
--- a/3_javascript/javascript1/week1/script2.js
+++ b/3_javascript/javascript1/week1/script2.js
@@ -1,0 +1,11 @@
+let firstWords = ["Eazy", "Awesome", "Corporate", "Innovative", "Global", "Dynamic", "Text", "Creative", "Fusion", "Spark"];
+let secondWords = ["Tech", "Solutions", "Labs", "Innovations", "Hub", "Network", "Edge", "Forge", "Nest", "Wave"];
+
+let randomIndexFirst = Math.floor(Math.random() * 10);
+let randomIndexSecond = Math.floor(Math.random() * 10);
+
+let startupName = firstWords[randomIndexFirst] + " " + secondWords[randomIndexSecond];
+let nameLength = startupName.length;
+
+console.log("The startup: " + startupName + " contains " + nameLength + " characters");
+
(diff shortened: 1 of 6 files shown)


3_javascript/javascript1/week1/script2.js
```diff
@@ -0,0 +1,11 @@
+let firstWords = ["Eazy", "Awesome", "Corporate", "Innovative", "Global", "Dynamic", "Text", "Creative", "Fusion", "Spark"];
+let secondWords = ["Tech", "Solutions", "Labs", "Innovations", "Hub", "Network", "Edge", "Forge", "Nest", "Wave"];
+
+let randomIndexFirst = Math.floor(Math.random() * 10);
+let randomIndexSecond = Math.floor(Math.random() * 10);
+
+let startupName = firstWords[randomIndexFirst] + " " + secondWords[randomIndexSecond];
+let nameLength = startupName.length;
```
SvitlanaBurlaka, line 8: If you want to know the exact number of characters, you can use`let nameLength = startupName.length - 1` . Because you are counting with the empty space between two words.

3_javascript/javascript1/week1/script3.js
```diff
@@ -0,0 +1,33 @@
...
+  
+  function comparePrices(expectedPrice, actualPrice, personName) {
+    if (expectedPrice > actualPrice) {
+      return personName + " is paying too much for the house.";
+    } else {
+      return personName + " isn't paying too much for the house.";
+    }
+  }
+  
+    let peterWidth = 8;
```
SvitlanaBurlaka, line 15: you've used let for all the variables. Since the dimensions of Peter's and Julia's houses and the sizes of their gardens are not expected to change, you could consider using const for these variables. This helps to convey the intention that these values should remain constant 😄

SvitlanaBurlaka (commented):
Good work ⭐  I want to recommend to pay attention and change from `let` to `const` your variables.  Always remember: by using const, you make it clear that these values should not be reassigned. If you later find a need to reassign a value, you can then consider whether it makes sense to use let. This can improve code readability and reduce the risk of accidental reassignments in future more complex work :)

PR #86, opened 2023-11-15
javascript1-week2/Daisy

Daisyangtan:


3_javascript/javascript1/week2/script1.js
```diff
@@ -0,0 +1,31 @@
+
+//Flight booking & Formal fullname function
+
+function getFullname(firstname, surname, useFormalName = false, isWoman = false) {
+    if (firstname === "" || surname === "") {
```
BrantalikP, line 5: Not crucial, but just to be sure, you can check whether the value is also a valid string to prevent inserting invalid types (number, undefined etc) as 

`if (typeof firstname !== 'string' || typeof surname !== 'string' || firstname === "" || surname === "") {`

3_javascript/javascript1/week2/script1.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+  
+    let fullName = firstname + " " + surname;
+  
+    if (useFormalName) {
+      if (isWoman) {
+        fullName = "Ms. " + fullName;
+      } else {
+        fullName = "Lord " + fullName;
+      }
```
BrantalikP, line 16: Bravo 🚀 I really love how clean the logic is.

3_javascript/javascript1/week2/script4.js
```diff
@@ -0,0 +1,44 @@
...
+function addStudentToClass(studentName) {
+  const maxStudents = 6;
+
+  if (class07Students.length >= maxStudents) {
+    console.log("Cannot add more students to class 07");
+    return;
+  }
+
+  if (class07Students.includes(studentName)) {
+    console.log('Student ${studentName} is already in the class');
```
BrantalikP, line 16: If you want to use variables inside of the string you need to use backquotes ` `` `
\`Student ${studentName} is already in the class\` otherwise you see 

`'Student ${studentName} is already in the class'` instead

3_javascript/javascript1/week2/script4.js
```diff
@@ -0,0 +1,44 @@
...
+    return;
+  }
+
+  if (studentName === "Queen") {
+    class07Students.push(studentName);
+    console.log("The Queen has a special place in class 07!");
+  } else if (studentName.trim() !== "") {
+    class07Students.push(studentName);
+    console.log('Student ${studentName} has been added to class 07');
```
BrantalikP, line 25: same here :) Use backquotes instead

3_javascript/javascript1/week2/script4.js
```diff
@@ -0,0 +1,44 @@
...
+    console.log("Cannot add more students to class 07");
+    return;
+  }
+
+  if (class07Students.includes(studentName)) {
+    console.log('Student ${studentName} is already in the class');
+    return;
+  }
+
+  if (studentName === "Queen") {
```
BrantalikP, line 20: You need to move this condition up. Otherwise, because of the return statement in this function, it won't reach the queen part

```
 if (class07Students.length >= maxStudents) {
    console.log("Cannot add more students to class 07");
    return;
  }
```

3_javascript/javascript1/week2/script5.js
```diff
@@ -0,0 +1,44 @@
...
+  const boughtCandyPrices = [];
+  let amountToSpend = Math.random() * 100;
+  
+  function addCandy(candyType, weight) {
+    const pricePerGram = candyPrices[candyType];
+    if (pricePerGram !== undefined) {
+      const totalPrice = pricePerGram * weight;
+      boughtCandyPrices.push(totalPrice);
+      amountToSpend -= totalPrice;
+      console.log('Added ${weight} grams of ${candyType} candy to your cart.');
```
BrantalikP, line 21: To allow using variables you need to use backquotes again :) Otherwise you will see this output `You need to move this condition up. Otherwise, because of the return statement in this function, it won't reach the queen.`

BrantalikP (changes requested):
Really great job! 👏 I left a few comments but I love how the code is structured and clean. This is a preference that is highly appreciated if you want to work as a Software Engineer, especially if you are working in a team.

PR #88, opened 2023-11-15
javascript1-week2/yousseftopaji

yousseftopaji:


3_javascript/javascript1/week2/js.js
```diff
@@ -0,0 +1,91 @@
+// Flight booking fullname function & Formal fullname
+
+
+function getFullname(firstname , surname, sex=false, useformalname=false){
+    if (useformalname && sex){
```
SvitlanaBurlaka, line 5: You can try to use here template literals , more info ▶️ [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals]().
For example : `return `Lady ${firstname} ${surname}`;`

3_javascript/javascript1/week2/js.js
```diff
@@ -0,0 +1,91 @@
...
+
+// Weather Wear 
+
+function temperatureOut(temperature){
+    if (temperature > 40){
+        return "Do not go outside if you don't want to be grilled"
+    }
+    else if (temperature > 20){
+        return "You should wear t-shirt and remember to apply sunscreen";
+    } else if (temperature < 20 && temperature > 15){
```
SvitlanaBurlaka, line 30: I would recommend to change this condition :
`} else if (temperature >= 15 && temperature <= 20) {
        return "You should wear a sweater";`

3_javascript/javascript1/week2/js.js
```diff
@@ -0,0 +1,91 @@
...
+}
+getNumberOfStudents();
+
+
+// Event Application
+const getEventDay = (daysUntilEvent) => {
+    const today = new Date();
+    const futureDate = today.getDay();
+    const weekdays = ["Monday","Tuesday","Wednesday","Thursday","Friday","Saturday","Sunday"];
+    const futureEvent = (futureDate + daysUntilEvent -1) % 7;
```
SvitlanaBurlaka, line 88: Just wondering, why you have here -1 in this piece of code ? 🤓 `const futureEvent = (futureDate + daysUntilEvent -1) % 7;`

SvitlanaBurlaka (commented):
Looks good, just added some minor comments! ⭐

3_javascript/javascript1/week2/js.js
yousseftopaji: I am wondering also about how if I do not do it like this it shows a day extra in the result -it returns Sunday instead of Saturday- and therefore the solution was either change the order of the array "weekdays" to start with Sunday instead of Monday, or to put -1 in the code line 88.

3_javascript/javascript1/week2/js.js
yousseftopaji: Thanks, this is useful.

3_javascript/javascript1/week2/js.js
yousseftopaji: What could be the alternative in this case?
And, is it a bad line?

yousseftopaji: or you mean it would be like 
else if (temperature > 15){
    return "You should wear a sweater";
}

SvitlanaBurlaka: Generally, your code here is right, it is me just being picky 🤓
But I would say that solution I proposed ensures that the condition is true when the temperature is 15, 16, 17, 18, 19, or 20,  correctly recommend what to wear in this range temperature. You will ensure here, that when it will be equal to 15 or 20, functionn will work right, in my point of view: :)

3_javascript/javascript1/week2/js.js
SvitlanaBurlaka: I found what is the trick here 😄 I also learned just now this : in `getDay()` return value will be an integer, between 0 and 6, representing the day of the week for the given date according to local time: 0 for Sunday, 1 for Monday, 2 for Tuesday, and so on. Check out this link : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay
Hence, you can have an array of dates, which starts from 'Sunday', then you do not need to use -1 in futureEvent, or leave it as you have. Code will be in both scenarios right, it is up to you what to use. 
Hope, you understood me what I meant 😄

PR #92, opened 2023-11-17
javascript1-week2/Monika

Monika-Marciniak:


3_javascript/javascript1/week2/homework2ndweek/fullname.html
```diff
@@ -0,0 +1,24 @@
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Flight booking</title>
+  </head>
+  <body>
+    <script>
+      const firstname1 = getFullName("Monika", "Marciniak");
```
MJanOthman: `firstName` and `lastName` are variables that should be passed to the function `getFullName` as paarmeters.

3_javascript/javascript1/week2/homework2ndweek/fullname.html
```diff
@@ -0,0 +1,24 @@
...
+      const firstname2 = getFullName("John", "Smith", false);
+
+      function getFullName(firstName, surName, isWoman = true) {
```
MJanOthman: is woman is another case, what you should implement first is `useFormalName`, so when `useFormalName= true` then you add Lord or Lady depending on the gender.

3_javascript/javascript1/week2/homework2ndweek/fullname.html
```diff
@@ -0,0 +1,24 @@
...
+        if (isWoman) {
```
MJanOthman: if you do not pass a value for firstName or surName, then you will get `undefined`
ex:
const firstname1 = getFullName("Marciniak") 
is Lady Marciniak undefined.
This should be fixed so it should not print `undefined`.

3_javascript/javascript1/week2/homework2ndweek/calendar.html
```diff
@@ -0,0 +1,36 @@
...
+        "Friday",
+        "Saturday",
+      ];
+
+      function getEventDay(day) {
+        const someDay = todaysWeekDay + day;
+        if (someDay > 7) {
+          const thatDay = someDay % 7;
+
+          console.log(weekDays[thatDay]);
```
MJanOthman: return instead of console log.

3_javascript/javascript1/week2/homework2ndweek/calendar.html
```diff
@@ -0,0 +1,36 @@
...
+        } else {
+          console.log(someDay);
```
MJanOthman: again, return the value instead of console log it

3_javascript/javascript1/week2/homework2ndweek/weather.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Weather</title>
+  </head>
+  <body>
+    <script>
+      function weather(temp) {
```
MJanOthman: you can be a bit more creative with the functions name.
the prop `temp` try to use `temperature` instead, the props should alaways be clear and meaningful.

3_javascript/javascript1/week2/homework2ndweek/student.html
```diff
@@ -0,0 +1,49 @@
...
+    <title>Student manager</title>
+  </head>
+  <body>
+    <script>
+      const class07Students = [];
+
+      function addStudentToClass(studentName) {
+        if (class07Students.includes(studentName)) {
+          return (
+            "Student " + "" + studentName + " is already in the class" + "."
```
MJanOthman: this is more easy to use
`Student ${studentName} is already in the class.`

MJanOthman (commented):
Great Effort so far, you should not have a separate html file for each task
and you should not implement the JS inside the HTML files.
The JavaScript should be written in a `.js` file.
What you did is not wrong, and you will mainly do it in react (in the near future), but for now keep the JS inside a `.js` file.

3_javascript/javascript1/week2/homework2ndweek/fullname.html
Monika-Marciniak: const firstname1 = getFullName(firstName, surName); Like that?

3_javascript/javascript1/week2/homework2ndweek/calendar.html
Monika-Marciniak: I did it with another mentor and thats what i was told to use, is there any specific reason why not to use console log? it doesnt work?

3_javascript/javascript1/week2/homework2ndweek/weather.html
Monika-Marciniak: temp is clear for me

3_javascript/javascript1/week2/homework2ndweek/student.html
Monika-Marciniak: its what was written in the task description 'Student Benjamin is already in the class', looks like string to me.

3_javascript/javascript1/week2/homework2ndweek/fullname.html
Monika-Marciniak: I dont exactly understand whats wrong, i think it works.

3_javascript/javascript1/week2/homework2ndweek/fullname.html
MJanOthman: I mean the const name should be `fullName1` and `fullName2`

3_javascript/javascript1/week2/homework2ndweek/calendar.html
MJanOthman: it works, and for now you are all fine.
I am sorry to make this a bit complicated, but in real-life coding, functions usually return values, they never console.log them.
What you did is 100% correct for now.

3_javascript/javascript1/week2/homework2ndweek/weather.html
MJanOthman: I know, and it is also clear to me as well.
The function is 100% correct.
But imagine you have a file with too many functions, and every function has too many props, and you have many teammates that work with you, then you should try to make everything clear for everyone.

3_javascript/javascript1/week2/homework2ndweek/student.html
MJanOthman: Sorry for the confusion, the value you are returning is 100% correct, I just tried to give you an easier way to do it without all the + and " "
check this:
https://www.shecodes.io/athena/8931-creating-a-string-with-variables-in-javascript

PR #94, opened 2023-11-18
javascript1-week1/olesia

AlesyaMazurenko:


helo.js
```diff
@@ -0,0 +1,123 @@
+// You are talking to a friend who suddently looks at you
```
BrantalikP, line 1: Is there any difference between the ﻿`exercise1` file and ﻿`helo.js`? If not, you can delete the duplicate file. :)

exersice1.js
```diff
@@ -0,0 +1,100 @@
+// You are talking to a friend who suddently looks at you 
+// and asks: "How old will you be in 2045?" Hmm you remember
```
BrantalikP, line 2: the file should be located in `/3_javascript/javascript1/week1` dir I think

exersice1.js
```diff
@@ -0,0 +1,100 @@
+// You are talking to a friend who suddently looks at you 
+// and asks: "How old will you be in 2045?" Hmm you remember 
+// the year you were born and try to do some calculation:
+// Born in 1987 + 3 that's 1990. 90 - 40 is 50 + 5 thats 58. 
+// I will be 58! Wow that was painful! Let's fix that by making 
+// some code that automatically does this!
+
+let yearOfBirth = 1986;
```
BrantalikP, line 8: The ﻿`let` keyword is used only if you expect that the variable will be changed, as it allows reassignment. However, for a constant variable like ﻿`yearOfBirth`, you should use the ﻿`const` keyword. Here is an example of using ﻿let:
```
let name = "petr"
console.log(name) // petr
name = lucas 
console.log(name) // lucas
```

exersice1.js
```diff
@@ -0,0 +1,100 @@
...
+let yearFuture = 2051;
+let age;
+
+age = yearFuture - yearOfBirth;
```
BrantalikP, line 12: This is an unnecessary step. :) You can assign the value as you declare the variable, like so:
 `
const age = yearFuture - yearOfBirth;
`

exersice1.js
```diff
@@ -0,0 +1,100 @@
...
+// The same friend(who by the way loves dogs) asks how old his 
+// dog will be in 2045. Hmm you think, lets make this into a
+// product as well! Dogs age can either be measured in dog years
+// or in human years, this we want to take into consideration!
+
+const dogYearOfBirth = 2022;
+const dogYearFuture = 2025;
+let dogYear;
+
+dogYear = dogYearFuture - dogYearOfBirth;
```
BrantalikP, line 26: again, you can assign the value beforehand :) 

`const dogYear = dogYearFuture - dogYearOfBirth;`

exersice1.js
```diff
@@ -0,0 +1,100 @@
...
+
+const shouldShowResultInDogYears = true;
+
+if (shouldShowResultInDogYears === false) {
+  console.log(`Your dog will be ${dogYear * 7} dog years old in ${dogYear}`);
+} else {
+  console.log(`Your dog will be ${dogYear} human years old in ${dogYear}`);
```
BrantalikP, line 33: I think there should be `dogYearFuture` instead of `dogYear` in the console log statements

exersice1.js
```diff
@@ -0,0 +1,100 @@
...
+// size.Lets help them figure out if they paid too much:
+
+const widthHouse = 8;
+const heightHouse = 10;
+const depthHouse = 10;
+const gardenSize = 100;
+
+//housePrice = volumeInMeters * 2.5 * 1000 + gardenSizeInM2 * 300;
+const peterHouseParameters = {width: 8, heigth: 10, deep: 10, garden: 100, price: 2500000}
+let housePricePeter = peterHouseParameters.width * peterHouseParameters.heigth* peterHouseParameters.deep * 2.5 * 1000 + peterHouseParameters.garden * 300;
```
BrantalikP, line 50: please use `const` keyword here :)

exersice1.js
```diff
@@ -0,0 +1,100 @@
...
+console.log("house Price Peter", housePricePeter);
+if (housePricePeter > peterHouseParameters.price) {
```
BrantalikP, line 52: I love the approach you have chosen! 👏 However, I believe the condition should be reversed, correct? If the real price (housePricePeter) is lower, then that should indicate that the price is too high.

exersice1.js
```diff
@@ -0,0 +1,100 @@
...
+  "Corporation",
+  "Sience",
+  "Steps",
+  "Alpha",
+  "Comp",
+];
+
+const randomNumber = Math.floor(Math.random() * 10);
+const newName = firstWords[randomNumber] + ' ' + secondWords[randomNumber];
+console.log(`The startup: "${newName}" contains ${newName.length} characters`)
```
BrantalikP, line 99: Nice! Just one thing, when you use ﻿`newName.length` it also counts the part ﻿` + ' ' +` in the length. If you want to count just the number of the letters, you should count them separately like this:

```
const firstWordCount = firstWords[randomNumber].length
const secondWordCount = secondWords[randomNumber].length
const numberOfLetters = firstWordCount + secondWordCount

console.log(`The startup: "${newName}" contains ${numberOfLetters} characters`)
```

or you can simply subtract 1 from the sum :D 
`
const numberOfLetters = newName.length - 1
`

BrantalikP (changes requested):
Great job @AlesyaMazurenko  👏 I really like how the code is clean and simple! Just left a few comments, but nothing crucial 🚀

exersice1.js
AlesyaMazurenko: You're right, but in our hometask  was an example that counts all symbols in new name.
  "An example could be: "The startup: "Easy Corporation" contains 16 characters" "

helo.js
AlesyaMazurenko: yes, its dublicates.

exersice1.js
BrantalikP: ok I see, in that case, all good:)

helo.js
BrantalikP: Then we can remove the file :)

PR #97, opened 2023-11-18
javascript1-week2/Anjali

anjali744:
Oh, homework is completed with the exercise. Thank you for challenging homework. :)

### DIFF
--- a/3_javascript/javascript1/week2/script.js
+++ b/3_javascript/javascript1/week2/script.js
@@ -78,3 +78,11 @@ addStudentToClass("Babul");
 addStudentToClass("Anshu");
 addStudentToClass("boo");
 addStudentToClass("queen");
+
+//student Manager-getNumberOfStudents
+
+function getNumberOfStudents() {
+  return class07Students.length;
+}
+
+//Candy helper optinal
--- a/3_javascript/javascript1/week2/script.js
+++ b/3_javascript/javascript1/week2/script.js
@@ -52,3 +52,29 @@ function getEventWeekday(weekOfTheEvent) {
   return daysOfWeek[targetWeek];
 }
 console.log("9 weeks from today will be a " + getEventWeekday(9));
+
+//student Manager-addStudentToClass
+const class07Students = [];
+function addStudentToClass(studentName) {
+  switch (true) {
+    case class07Students.includes(studentName):
+      console.log("Student ${studentName} is already in the class.");
+      break;
+    case class07Students.length >= 6 && studentName.toLowerCase() !== "queen":
+      console.log("Cannot add more students to class 07.");
+      break;
+    default:
+      class07Students.push(studentName);
+      console.log("Student ${studentName} has been added to class 07.");
+      break;
+  }
+}
+
+addStudentToClass("Rishab");
+addStudentToClass("Jyansh");
+addStudentToClass("Mohit");
+addStudentToClass("Rooney");
+addStudentToClass("Babul");
+addStudentToClass("Anshu");
+addStudentToClass("boo");
+addStudentToClass("queen");
--- a/3_javascript/javascript1/week2/script.js
+++ b/3_javascript/javascript1/week2/script.js
@@ -65,7 +65,7 @@ function addStudentToClass(studentName) {
       break;
     default:
       class07Students.push(studentName);
-      console.log("Student ${studentName} has been added to class 07.");
+      console.log(`Student ${studentName} has been added to class 07.`);
       break;
   }
 }
@@ -79,10 +79,7 @@ addStudentToClass("Anshu");
 addStudentToClass("boo");
 addStudentToClass("queen");
 
-//student Manager-getNumberOfStudents
-
 function getNumberOfStudents() {
   return class07Students.length;
 }
-
-//Candy helper optinal
+console.log(class07Students);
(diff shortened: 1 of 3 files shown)


3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,85 @@
+//Flight Booking Fullname
+function fullName(firstName, surName) {
```
MJanOthman, line 2: You do not need this function anymore, the second function `getFullname` handle all the cases.

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,85 @@
...
+const fullName1 = fullName("Anjali", "Kumari");
+const fullName2 = fullName("Babul", "Raja");
+console.log(fullName1);
+console.log(fullName2);
+
+//Formal Name
+function getFullname(firstName, surName, useFormalName, isWoman) {
+  if (useFormalName) {
+    return "Lord" + " " + firstName + " " + surName;
+  } else if (isWoman) {
```
MJanOthman, line 14: The condition  `if(isWoman)` should be nested inside the first condition `if(useFormalName)`.
according to your code, the following cases happen:

```suggestion
  } const formalName1 = getFullname("Anjali", "Kumari", false, true);//1 Miss Anjali Kumari
const formalName2 = getFullname("Anjali", "Kumari", true, true);//2 Lord Anjali Kumari
const formalName3 = getFullname("Anjali", "Kumari", false, false);//3 Anjali Kumari
const formalName4 = getFullname("Anjali", "Kumari", true, false);//4 Lord Anjali Kumari {
```
The first and second one are wrong.

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,85 @@
...
+  } else {
+    return firstName + " " + surName;
+  }
+}
+const formalName = getFullname("Anjali", "Kumari", false, true);
+console.log(formalName);
+//note: other way to solve the promblem is = isWoman ? 'Lady' : 'Lord';
+
+//Weather Wear
+function todayWeatherTemp(temp) {
```
MJanOthman, line 25: giving a short name for the props is a good idea, but sometimes it is important to give a good and understandable name
This comment is not to tell that what you did is wrong, it is just to let you do it in a better way.

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,85 @@
...
+}
+console.log(todayWeatherTemp(37));
+
+//Event Application
+
+const todayDay = new Date();
+const todayWeek = todayDay.getDay();
+const targetWeek = (todayWeek + 9) % 7;
+
+function getEventWeekday(weekOfTheEvent) {
```
MJanOthman, line 42: you are not using the prop you passed `(weekOfTheEvent)` anywhere inside the function.
you depends on 3 const outside this function, implement them inside the function and refactor it.

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,85 @@
...
+  return daysOfWeek[targetWeek];
+}
+console.log("9 weeks from today will be a " + getEventWeekday(9));
+
+//student Manager-addStudentToClass
+const class07Students = [];
+function addStudentToClass(studentName) {
+  switch (true) {
+    case class07Students.includes(studentName):
+      console.log("Student ${studentName} is already in the class.");
```
MJanOthman, line 61: the output for this case is `Student ${studentName} is already in the class.`
nice that you know this ${studentName} trick, but to make it works you need to use `` instead of ""

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,85 @@
...
+    "Friday",
+    "Saturday",
+  ];
+  return daysOfWeek[targetWeek];
+}
+console.log("9 weeks from today will be a " + getEventWeekday(9));
+
+//student Manager-addStudentToClass
+const class07Students = [];
+function addStudentToClass(studentName) {
```
MJanOthman, line 58: The conditions work as they should, very nice work, but there is a point you should be aware of:
switch case should return instead of console.log
so return the values inside the function, and then do the console.log when you are calling the function.


```suggestion
function addStudentToClass(studentName) {
     switch (true) {
    case class07Students.includes(studentName):
      return("Student ${studentName} is already in the class.");
      break;
    case class07Students.length >= 6 && studentName.toLowerCase() !== "queen":
      return("Cannot add more students to class 07.");
      break;
    default:
      class07Students.push(studentName);
      return(`Student ${studentName} has been added to class 07.`);
      break;
  }
}
 console.log(addStudentToClass("Rishab"))
 console.log(addStudentToClass("Jyansh"))
.
.
.

```

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,85 @@
...
+addStudentToClass("Rooney");
+addStudentToClass("Babul");
+addStudentToClass("Anshu");
+addStudentToClass("boo");
+addStudentToClass("queen");
+
+function getNumberOfStudents() {
+  return class07Students.length;
+}
+console.log(class07Students);
```
MJanOthman, line 85: nit:
also console.log(getNumberOfStudents())

MJanOthman (commented):
Great job, well done
some comment are left below 👇

PR #98, opened 2023-11-19
javascript1-week2/Ihor

IhorKulytskyi:
Finished my homework JS-week2

### DIFF
--- a/3_javascript/javascript1/week2/script.js
+++ b/3_javascript/javascript1/week2/script.js
@@ -1,0 +1,93 @@
+// getFullname
+
+function getFullname(firstname, surname, useFormalName, gender) {
+  if (useFormalName) {
+    return "Lord" + " " + firstname + " " + surname;
+  } else if (gender) {
+    return "Lady" + " " + firstname + " " + surname;
+  } else {
+    return firstname + " " + surname;
+  }
+}
+
+const fullname1 = getFullname("Ihor", "Kulytskyi");
+const fullname2 = getFullname("Benjamin", "Hughes");
+const fullname3 = getFullname("Maria", "Magdalena", false, true);
+
+console.log(fullname1);
+console.log(fullname2);
+console.log(fullname3);
+
+//getEventWeekday
+
+function getEventWeekday(dayEvent) {
+  let today = new Date();
+  let weekdays = [
+    "sunday",
+    "monday",
+    "tuesday",
+    "wednesday",
+    "tursday",
+    "freday",
+    "saturday",
+  ];
+  let indexDay = (today.getDay() + dayEvent) % 7;
+  return weekdays[indexDay];
+}
+console.log(getEventWeekday(3));
+
+// Weather wear
+
+function checkTemperature(temperature) {
+  if (temperature <= 5 && temperature >= 0) {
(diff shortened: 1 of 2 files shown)


3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,93 @@
+// getFullname
+
+function getFullname(firstname, surname, useFormalName, gender) {
+  if (useFormalName) {
+    return "Lord" + " " + firstname + " " + surname;
+  } else if (gender) {
+    return "Lady" + " " + firstname + " " + surname;
+  } else {
+    return firstname + " " + surname;
```
margaritluch, line 9: What if the surname is not present? What if I call the function like this `getfullName("Margarit")`? 🤔

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,93 @@
+// getFullname
+
+function getFullname(firstname, surname, useFormalName, gender) {
+  if (useFormalName) {
+    return "Lord" + " " + firstname + " " + surname;
```
margaritluch, line 5: I recommend using template literals instead. They make the code more readable and elegant.[ Here](https://www.w3schools.com/js/js_string_templates.asp) you can read more about it  📃😊

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,93 @@
...
+
+console.log(fullname1);
+console.log(fullname2);
+console.log(fullname3);
+
+//getEventWeekday
+
+function getEventWeekday(dayEvent) {
+  let today = new Date();
+  let weekdays = [
```
margaritluch, line 25: Weekdays variable can be const as its value is not going to change 🙃

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,93 @@
...
+  } else if (temperature <= 18 && temperature >= 15) {
+    console.log("Light jacket");
+  } else if (temperature <= 25 && temperature >= 19) {
+    console.log("Shorts and T-shirt");
+  } else if (temperature > 30) {
+    console.log("Don`t forget sunscreen");
+  } else {
+    console.log("Don`t go out, it`s cold there!");
+  }
+}
```
margaritluch, line 55: good use of else if 👍

3_javascript/javascript1/week2/script.js
```diff
@@ -0,0 +1,93 @@
...
+    if (class07Students[i] === studentName) {
+      console.log(`${studentName} in the class`);
+      return;
+    }
+  }
+  if (class07Students.length <= 6) {
+    class07Students.push(studentName);
+  } else if (studentName === "Queen") {
+    console.log("Welcome Queen of Denmark Margrethe II");
+  } else {
```
margaritluch, line 73: The text will be logged but if the number of students is bigger than 6 then the queen's name won't be added to the array 😬. To fix this issue you need to shrink the array size when the Quenn is added to a full class 🔧

margaritluch (commented):
Alt Text
Good job Ihor 👏 Keep up the motivation 🧠

PR #100, opened 2023-11-19
javascript1-week2/olesia

AlesyaMazurenko:


3_javascript/javascript1/week2/week2.js
```diff
@@ -0,0 +1,148 @@
...
+        return ('Lord ' + firstname + " " + surname);
+
+    } else {
+        return (firstname + " " + surname);
+    } 
+}  
+
+const fullname1 = getFullname("Benjamin", "Hughes", true);
+const fullname2= getFullname("Benjamin", "Hughes", false);
+
```
BrantalikP, line 21: great job, but I see that your solution lacks two more parts :)

1. `Remember to consider someone calling the function with an empty string as firstname and lastname.
`


2. `But what if the person is a woman? Describe how you would fix the getFullname so it also works for women`

3_javascript/javascript1/week2/week2.js
```diff
@@ -0,0 +1,148 @@
...
+// Event application
+
+function getEventWeekday(eventy) {
+    const weekday = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"]
+    const dateNow = new Date();
+    
+    const numberDayNow = dateNow.getDay();
+ 
+    const nextDate = (numberDayNow + eventy) % 7;
```
BrantalikP: `eventy` typo :)

3_javascript/javascript1/week2/week2.js
```diff
@@ -0,0 +1,148 @@
...
+        case 'not very cold': clothes= 'only jacket';
+        break;
+        case 'warm': clothes= 'jeans suit';
+        break;
+        case 'hot': clothes= 'shorts and a t-shirt';
+        break;
+        case 'very hot': clothes= 'do not forget swimming suit';
+        break;
+        default: clothes = 'do not forget wear shoes'
+    }
```
BrantalikP, line 69: Nice! I like the structure, but I think that one of the condition is redundant, you can already assign the clothes value in the first condition I can see :) 

```
if (temp < 0) {
    weather = 'cold'
    clothes= 'down jacket and hat';
  } else if (temp >= 0 && temp <= 7) {
    weather = 'not very cold'
    clothes= 'only jacket';
  } else if (temp > 7 && temp <= 15) {
    weather = 'warm'
    clothes = 'jeans suit'
  } else if (temp > 15 && temp <= 25) {
    weather = 'hot'
    clothes= 'shorts and a t-shirt';
  } else {
    weather = 'very hot'
    clothes= 'do not forget swimming suit';
  }
```

3_javascript/javascript1/week2/week2.js
```diff
@@ -0,0 +1,148 @@
...
+        console.log("Cannot add more students to class 07");
+      }
+    } else {
+        console.log("Cannot add empty name");
+    }
+}
+
+function getNumberOfStudents() {
+    return class07Students.length;
+  // You write code here
```
BrantalikP: you can remove this comment :)

3_javascript/javascript1/week2/week2.js
```diff
@@ -0,0 +1,148 @@
...
+addStudentToClass("N2");
+addStudentToClass("N3");
+addStudentToClass("N1");
+addStudentToClass('Queen4');
+addStudentToClass("Queen");
+addStudentToClass("");
+addStudentToClass("N6");
+addStudentToClass("N7");
+console.log(class07Students);
+console.log(`The number of students is ${class07Students.length}`);
```
BrantalikP: I think you are supposed to get the length via `getNumberOfStudents` function, try to call it inside the console log

3_javascript/javascript1/week2/week2.js
```diff
@@ -0,0 +1,148 @@
...
+let boughtCandyPrices = [];
+const typeCandy = ["sweet", "chocolate", "toffee", "chewing-gum"];
+const priceCandy = [0.5, 0.7, 1.1, 0.03];
+const amountToSpend = Math.random() * 100;
+
+console.log("amountToSpend", amountToSpend);
+
+function addCandy(candyType, weight) {
+    let price = 0;
+    for (i = 0; i < typeCandy.length; i++) {
```
BrantalikP: its a valid solution for sure :) but since there is a for loop, the function would be called always 4times despite that `candyType` index. For this cases is much better to use object

```
const candys = {
  sweet: 0.5,
  chocolate: 0.7,
  toffee: 1.1,
  'chewing-gum': 0.03,
}

function addCandy(candyType, weight) {
  const price = candys[candyType] * weight
  boughtCandyPrices.push(price)
}`
```

3_javascript/javascript1/week2/week2.js
```diff
@@ -0,0 +1,148 @@
...
+    while (i < boughtCandyPrices.length) {
+        totalSum =+ boughtCandyPrices[i];
+        if (totalSum < amountToSpend) {
+            console.log("You can buy more, so please do!");
+        } else {
+            console.log("Enough candy for you!");
+        }
+        i++;
+    }
+}
```
BrantalikP, line 153: Great job! the task is not easy at all, but you made it. 

Few notes
1.   the if else condition should be after you finish your loop logic
```
  if (totalSum < amountToSpend) {
            console.log("You can buy more, so please do!");
        } else {
            console.log("Enough candy for you!");
        }
```
        
2. `while`  is valid for sure, but it can be a bit tricky something (or even dangerous!). Here is an alternative approach :)
```
  let totalSum = 0

  for (const price of boughtCandyPrices) {
    console.log({ price })
		totalSum += price
  }

  if (totalSum < amountToSpend) {
    console.log('You can buy more, so please do!')
  } else {
    console.log('Enough candy for you!')
  }

```

BrantalikP (changes requested):
Although the tasks are getting harder, you are doing a great job 👏 Please check out the comments and let me know if you have any questions :)

3_javascript/javascript1/week2/week2.js
AlesyaMazurenko: yes, i've want tried to use 'switch'

3_javascript/javascript1/week2/week2.js
```diff
@@ -0,0 +1,155 @@
+// Flight booking fullname function
+
+function getFullname(firstname, surname, useFormalName, isLady) {
+    if (firstname && surname) {
```
BrantalikP, line 4: you still need to check whether the string is empty or not :) You can find the solution [here](https://www.freecodecamp.org/news/check-if-string-is-empty-or-null-javascript/), if not. Please reach out to me and I can help you with that:)

3_javascript/javascript1/week2/week2.js
BrantalikP: okok :) But also important is to make code as clean as possible, therefore, it is better to use just one check if its possible  (e.g my example above), anyway good job!:)

BrantalikP (approved):

PR #101, opened 2023-11-19
javascript1-week2/Helena

helena69hme:


3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
+//Welcome to my second experience in JAVASCRIPT :)
+
+/* console.log("*******     Flight booking fullname function    *******");
```
MJanOthman: This comment is not explaining anything, you do not need it.

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+ const fullname1 = getFullname("Helena", "Hashemi");
+ const fullname2 = getFullname("Shahram", "Shokri");
+
+ console.log(fullname1);
+ console.log(fullname2); */
+
+console.log("*******     Flight booking fullname function / ('Formal fullnam')     *******");
+function getFullname(firstname, surname, gender, userFormalName) {
```
MJanOthman: You passed a parameter `userFormalName` (and that is perfect) but you are not using it.

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
...
+
+    if(gender === "male") {
```
MJanOthman: here you should check for `userFormalName` before you check the gender, think of it this way:
 if I want to use formal name => if the gender is `male` ,I will use `Lord`, else if the gender is `female` I will use `Lady`

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
...
+}
+if(gender === "female"){
+    return `Lady ${firstname} ${surname}`;
+}
+if (gender === ""){
+    return `${firstname} ${surname}`;
+}
+}
+
+const userFormalName = getFullname("Soeren", "Sepehr", "");
```
MJanOthman: if you correct the code above, then you nedd to pass true or false, depending on what value you want to assign to `userFormalName`

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
...
+//         return `Frazing...! Wear ${wearInFrazing}.`
+//     }
+//     if(tempreture > 10 && tempreture < 19) {
+//         return `Almost cold...! Wear ${wearInCold}.`
+//     } else {
+//         return `Sun shine...! Wear ${wearInWarm} and remember suncream and sunglasses.`
+//     }
+// }
+// const wearInWarm = outsideTempreture(27);
+// console.log(wearInWarm);
```
MJanOthman: This function that is commented out is the rught answer!

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
...
+
+const tempreture = 20;
+switch(tempreture) {
```
MJanOthman: this is not a function, you are just creating a switch case, but it is not implemented inside a function.

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
...
+
+    default:
+    console.log("Warning!... Risk of heatstroke");   
+}
+
+//In this case there is some problem. As you can see I have set the temprature on 20:
+//here (const tempreture = 20;). So we should face with this sentence:
+//(Sun shine...! Wear top, shorts and remember suncream and sunglasses.)
+//but we have (Warning!... Risk of heatstroke) in the browser and I cant undrestand what is wrong with that :(
+//please give me a clue!
```
MJanOthman: The clue is to not use switch case, use for loop instead.
the reason is, whatever type of variable you use in switch, you should use the same in the cases, so:

this should be ok
```suggestion
temprature1 = 20;
switch (temprature1) {
  case 20:
    console.log("this is cold");
}
```
because temprature is a number, and the case is a number
but what you did is:

```suggestion
temprature1 = 20;
switch (temprature1) {
  case temprature===20:
    console.log("this is cold");
}
```
this does not work, because the case temprature===20 is a boolean not a number.

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
...
+
+console.log("*******     Weather wear     *******");
+const class07Students = [];
+function addStudentToClass(studentName) {
+    if(studentName === "") {
+        console.log("Add your name.");
+    } else /*if(studentName = Number){ 
+        console.log("Invalid request");
```
MJanOthman: remove the commented out code, it is always important to keep the code clean.

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,146 @@
...
+        class07Students.push(studentName);
+        console.log(`${studentName} added to the class.`);
+    } else { 
+        console.log("Error");
+    }
+}
+
+function getNumberOfStudents(numberOfStudent){
+return class07Students;
+}
```
MJanOthman: this is not the correct function to calculate number of students in the class.
you should pass the array of students as parameter, and return the length of the array.

MJanOthman (commented):
Very good work Helena, small points to think about again, and some cleaning for the code.

3_javascript/javascript1/week2/javascript.js
helena69hme: done!

3_javascript/javascript1/week2/javascript.js
helena69hme: I couldn't understand what exactly you mean, but tried to correct it

3_javascript/javascript1/week2/javascript.js
helena69hme: good explanation and thank you for reminding me to use "IF" "ELSE" :)
done!

3_javascript/javascript1/week2/javascript.js
helena69hme: Done!

3_javascript/javascript1/week2/javascript.js
helena69hme: I tried to show you how I think about answering the question, then I started to add the answers by commenting one of them.

3_javascript/javascript1/week2/javascript.js
helena69hme: I see

3_javascript/javascript1/week2/javascript.js
helena69hme: oooooh!
Now I understand.

super explanation! I appreciate it

3_javascript/javascript1/week2/javascript.js
helena69hme: Actually, that was my question.
 I wanted to make a condition by this subject:
 if(studentName === Number){ 
        console.log("Invalid request");
    } else

addStudentToClass(25);

But what I received in the browsers was:
"25 added to the class."

would you please tell me why it didn't work?!!

PR #106, opened 2023-11-21
javascript1-week2/anastasiia

humenyuk16:


3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,99 @@
+//Flight booking fullname function
+
+function getFullname(firstname, surname, gender, useFormalName = false) {
+    if (!firstname.trim() && !surname.trim()) {
+        return "Invalid name";
+    }
+    if (gender === "male") {
```
SvitlanaBurlaka, line 7: I think, to be more precise here, you would need to wrap your if's into one more if, where you will check if 
`useFormalName` is true or false :
```

  if (useFormalName) {
    if (gender === "male") {
      return "Lord " + firstname + " " + surname;
    } else if (gender === "female") {
      return "Lady " + firstname + " " + surname;
    } else {
      return firstname + surname;
    }
  }
```

3_javascript/javascript1/week2/javascript.js
```diff
@@ -0,0 +1,99 @@
...
+
+function getEventWeekday(daysToAdd){
+    const daysOfWeek = ["Sunday", 
+    "Monday", 
+    "Tuesday", 
+    "Wednesday", 
+    "Thursday", 
+    "Friday", 
+    "Saturday"];
+    const today = new Date().getDay();
```
SvitlanaBurlaka, line 34: good use of new Date and getDay methods 👍

SvitlanaBurlaka (commented):
You are doing great! I like how you solved it ❤️

PR #107, opened 2023-11-23
javascript1-week3/hadis

Hadis-jamali:


3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+    days: 2,
+    hours: 2,
+    minutes: 0,
+  },
+];
+
+let totalTimeMinutes = 0;
+
+for (const item of seriesDurations) {
+  const { days, hours, minutes } = item;
```
MaherHussain: good job . i like how you extract properties from object :-)

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+const totalMinutesInLifespan = 80 * 365 * 24 * 60;
+
+const percentageOfLifespan = (totalTimeMinutes / totalMinutesInLifespan) * 100;
+
+console.log(`Total time spent watching TV series: ${totalTimeMinutes} minutes`);
+console.log(
+  `Percentage of an 80-year lifespan spent watching TV series: ${percentageOfLifespan.toFixed(2)}%`
+);
+
+//NOnoN0nOYes (Note taking app)
```
MaherHussain: What is required in  this task is to create function. it's better all this stuff in function

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+const notes = [];
+
+function saveNote(content, id) {
+  const myWork = { id: id, content: content };
```
MaherHussain: choose better name for constant 
for example her should be 
`const note = { id: id, content: content };`

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+  notes.push(myWork);
+}
+saveNote("Pick up groceries", 1);
+saveNote("Do laundry", 2);
+console.log(notes);
+
+//Get a note
+function getNote(id) {
+  for (const item of notes) {
```
MaherHussain: let's call it `for (const note of notes)`

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+    if (item.id === id) {
+      return item;
+    }
```
MaherHussain, line 98: what if the id is not exist, and what if the user provide not number  id, you should consider all these cases

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+  }
+}
+
+const firstNote = getNote(1);
+console.log(firstNote);
+
+//Log out notes
+
+function logOutNotesFormatted(id) {
+  for (const item of notes) {
```
MaherHussain: same here use note in stead of item

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+}
+
+const firstNote = getNote(1);
+console.log(firstNote);
+
+//Log out notes
+
+function logOutNotesFormatted(id) {
```
MaherHussain: why you need parameter in this function

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+  for (const item of notes) {
+    if (item.id === id) {
```
MaherHussain: you do not need to check the id, and you need the id at all here. 
what we want in this function is to return all notes in formatted string

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,139 @@
...
+}
+
+logOutNotesFormatted();
+
+//CactusIO-interactive (Smart phone usage app) optional
+
+const activities = [];
+
+function addActivity(date, activity, duration) {
+  const item = { date: date, activity: activity, duration: duration };
```
MaherHussain: choose always you variable names that make sense to your task . it's good to learn that

PR #109, opened 2023-11-24
javascript1-week3/katayoun

katayouny:
Hi
I have done class exercises + some extra exercises + week 3 homework. 
I have written some questions inside the js codes, which I would appreciate if one of dear mentors could guide me and answer my questions, thank you.


3_javascript/javascript1/week3/excercisescript.js
```diff
@@ -0,0 +1,144 @@
+
+// __________________ The Last Class Excercises ______________________
+
+
```
BrantalikP, line 4: Hey, could you please remove all the files that are not related to the Homework, please :)
I was not sure at first what I should review :D

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
+// ________________warmup excercise : The time it will take to arrive at your destination _________________
+
+function travelInformation(speedInkmperh, destinationDistanceInkm) {
+
+    let timeToRichToTheDestination = (destinationDistanceInkm / speedInkmperh);
+    const hours = Math.trunc(timeToRichToTheDestination); //same result with const hours (~~timeToRichToTheDestination) to cuts all fractional digits
+    let minutes = (timeToRichToTheDestination - hours) * 60;
+    const rminutes = Math.ceil(minutes); // To round up the minutes
+    return ('Your travel takes ${hours} hours & ${rminutes} minutes'); // It doesn't give a result for ${hours} & ${rminutes} !!!!!
```
BrantalikP: Good job!

Only in order to  use `String interpolation` you must use backquotes `` instead of single-quotes '' :)

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+
+const travelTime = travelInformation(50, 432);
+console.log(travelTime);
+
+
+// ______ warmup excercise : how much time a tv series have taken as a percentage of an average lifespan of 80 years_______
+
+function logOutSeriesText(TVSeriestiltle, daysToWatch) {
+
+    const tookDaysOfLife = (80 * 365) / daysToWatch;
```
BrantalikP: Since you want to compare the watch duration against 80 years, you don't need to divide the number by ﻿`daysToWatch` here.

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+        title: "The Wire",
+        days: 27,
+        hours: 12,
+        minutes: 0,
+    },
+];
+
+
+for (let i = 0; i < seriesDurations.length; i++) {
+    let result = logOutSeriesText(seriesDurations[i].title, seriesDurations[i].days);
```
BrantalikP, line 53: Great start! You're almost there, but if you notice, you also have values for `hours` and `minutes` that need to be used. To find the accurate output, you need to use all of these values and then compare it against 80 years. To do this, you must transform all values to the same unit (hint: minutes 👀), and then you can use the formula above :)

Also you can use `.toFixed(3)` method to round the number to 3 decimal places

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+// _________Homework : Smart-ease / NOnoN0nOYes (Note taking app) / create a simple and easy notepad__________
+
+// ---------Save a note  ---------------------------------
+
+const notes = [];
+
+function saveNote(content, id) { // content parameter is a string and id is a number
+    let note = { content: content, id: id };
+    notes.push(note);
+    return notes;
```
BrantalikP: there is no need for `return` statement here, you can delete it :)

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+saveNote("Washing dishes", 3);
+saveNote("Ironing clothes", 4);
+console.log(notes); // [{content: 'Pick up groceries', id: 1}, {content: 'Do laundry', id: 2}, ........]
+
+
+//------  Get a note  ----------------------------------
+
+function getNote(id) {
+
+    if (id <= notes.length || typeof id !== 'string') {
```
BrantalikP: since `id`is not associated with the array length, you can remove this part `id <= notes.length || `
if the Notes would be in this format, the function would not work 
```
[
{content: 'someContent1", id: 1},
{content: 'someContent2", id: 2},
{content: 'someContent3", id: 3},
{content: 'someContent485", id: 485},
]
```
The length would be still 4 but the id 485 is bigger than that, see? :) 

also the `id` is only number hence its better to streamline the condition like this `typeof id === 'number'`

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+        return (noteT.content); //f.x  id=1 => "Pick up groceries"
+    } else {
+
+        return (alert(`Are you kidding me? :O A number !!
+        Run the program again and enter a number between 1 and ${notes.length}`));
+        //Why I get 'Undefind' after this alert message? :/
+    }
+};
+
+const getId = prompt(`Enter a number between 1 and ${notes.length} `, 1);
```
BrantalikP, line 93: Love that you used a prompt here! 👏

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+
+gottenDate = prompt('add a date fx. 23/7-2023', '23/7-2023');
+gottenActivity = prompt('add an activity fx. Instagram', 'Instagram');
+gottenDuration = prompt('Enter a number as activity duration hours fx.', 40);
+addActivity(gottenDate, gottenActivity, gottenDuration);
+
+gottenDate = prompt('add a date fx. 23/7-2023', '23/7-2023');
+gottenActivity = prompt('add an activity fx. Facebook', 'Facebook');
+gottenDuration = prompt('Enter a number as activity duration hours fx.', 15);
+addActivity(gottenDate, gottenActivity, gottenDuration);
```
BrantalikP, line 148: Again, using the prompt is really creative! :)

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+
+//PART1: Adding an activity ______________________________________
+
+const activities = [];  // [{}]  or [{}] would also work?
+
+function addActivity(date, activity, duration) {  //date: string, activity: string; duration: minutes
+
+    let activitiesDataDuration = { activity: activity, date: date, duration: duration };
+    activities.push(activitiesDataDuration);
+    return activities;
```
BrantalikP: no need for `return` statement. please delete the line :)

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+
+// -----------CactusIO-interactive (Smart phone usage app) optional -------------------
+
+//PART1: Adding an activity ______________________________________
+
+const activities = [];  // [{}]  or [{}] would also work?
+
+function addActivity(date, activity, duration) {  //date: string, activity: string; duration: minutes
+
+    let activitiesDataDuration = { activity: activity, date: date, duration: duration };
```
BrantalikP: a little trick to save some time and lines, 
if the `variable` and the object `key` are the same you can write it just once like this 
`let activitiesDataDuration = { activity, date, duration };`

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+
+// PART2: Adding an activity ____________________________________
+
+function showStatus(activities) {
+
+    if (activities.length === 0) {
+        let NoactivitiesAdded = console.log(`You have not added any activities. 
+        Add some activities before to request seeing the status of activities`);
+        //Should I call the function again???
+        return NoactivitiesAdded;
```
BrantalikP: you can simply call the console.log here and remove the return statement 

```
console.log(`You have not added any activities. 
        Add some activities before to request seeing the status of activities`);
else {

```

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+
+    } else {
+        console.log(`You have added ${activities.length} activities`);
+        for (let i = 0; i < activities.length; i++) {
+            console.log(`Activities ${i + 1}: ${activities[i].activity}`);
+        }
+
+        let totalSmartphoneUsagedurationInMinuits = 0;
+        for (let j = 0; j < activities.length; j++) {
+            totalSmartphoneUsagedurationInMinuits = totalSmartphoneUsagedurationInMinuits + activities[j].duration;
```
BrantalikP: Since the value from the prompt is always a string, adding the value will result in a different output than you might expect. :) You can check it for yourself :) Instead of the default ﻿`85`, the value will be ﻿`0304015`. To fix this, you need to cast the 'duration' value to a Number type. You can achieve this by either using `﻿Number(activities[j].duration)` or simply adding a ﻿`+` sign as in ﻿`+activities[j].duration.`

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+            //Why it gives a different digit whenever I run the codes???
+        }
+        console.log(`Activities amount to ${totalSmartphoneUsagedurationInMinuits} min. of usage on ${activities[0].date}`);
```
BrantalikP: If I understand correctly the activities can have a different date each, if so, using `${activities[0].date}` can give you an incorrect result. I recommend to remove the whole `of usage on date` part since its not part of the assignment anyway :)

3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,217 @@
...
+
+
+//___just a little class excercise ________________________________________________________
+
+/*
+const birds = ["Parrot", "Falcon", "Owl"];
+
+for (const bird of birds) {
+console.log(bird);
+};*/
```
BrantalikP: Lets remove the unused lines :)

BrantalikP (changes requested):
Great job! Almost there, please check out my comments, if you have any questions do not hesitate to reach out to me! You got this!

3_javascript/javascript1/week3/excercisescript.js
katayouny: Hi, I have also done some class exercises, which I wanted to ask to be reviewed, as there are also some issues there and some questions about codes ( inside excercisescript.js) , thank you :)

3_javascript/javascript1/week3/script.js
katayouny: Thank you. Done :)

3_javascript/javascript1/week3/script.js
katayouny: Thank you. Done :)

3_javascript/javascript1/week3/script.js
katayouny: Thank you.
I am still a little confused regarding my question about how I can add the 3 percentages and make a subtotal of them :).
I wanted to use a counter, however, I couldn't figure it out where to put that. Because when I want to return the results, I can return only to values. I don't know in which way I can also return the summation.

3_javascript/javascript1/week3/script.js
katayouny: Thank you, done.

3_javascript/javascript1/week3/script.js
katayouny: Thank you, done :)

3_javascript/javascript1/week3/script.js
katayouny: Thank you :)
Could you please tell me why the codes between lines 78 to 95 do not work properly? 
And why do I get 'Undefined' after this alert message?

3_javascript/javascript1/week3/script.js
katayouny: Thank you :)

3_javascript/javascript1/week3/script.js
katayouny: Thank you, done :)

3_javascript/javascript1/week3/script.js
katayouny: Thank you :)

3_javascript/javascript1/week3/script.js
katayouny: Thank you, done :)

3_javascript/javascript1/week3/script.js
katayouny: Thank you, done :).

3_javascript/javascript1/week3/script.js
katayouny: Thank you :)
The dates are the same, as we should find the usage duration for the predefined date. I have changed ${activities[0].date} it to ${activities.date} and now it works correctly.

3_javascript/javascript1/week3/script.js
katayouny: Thank you, done :).

3_javascript/javascript1/week3/script.js
BrantalikP: hmm, you can just print out  `totalPercentOfLife ` no?

3_javascript/javascript1/week3/script.js
BrantalikP: yes :)  if (typeof id === `number`) issue is here, if you want to refer to a type, you must to use quotes :) Otherwise it will look for a variable `number` (e.g const number) but there isn't such a variable so it throws an error :)

BrantalikP: Also, you must first parse the ﻿`id` to a number because all input values from the prompt come as a ﻿`string`.
To do so, you can either use casting as: ﻿`Number(id),` or simply add a ﻿`+` before it like: ﻿`+id`. Both approaches are valid. This only works on numbers within strings. If you try to give the ﻿id a value of "my-id", it won't work. That scenario would be caught by the else case, which is the correct behavior.

3_javascript/javascript1/week3/excercisescript.js
```diff
@@ -0,0 +1,144 @@
...
+    price: "35 kr."
+}
+
+pizza.ingredients = ["peperoni", "pepper", "tomato", "cheez", "onion", "mashrooms"];
+
+let key = prompt("Do you want corn as topping on your pizza? \n Plz answer with Yes or No");
+// "error" "ReferenceError: topping is not defined   ????????????
+
+// access by variable
+alert(pizza[key]);   //?????
```
BrantalikP, line 73: not sure wht is the goal of this exercise, but this will throw an undefined if the user answers either "yes" or "no" 

because the key will have a value "true" lets say, so you are alerting

`pizza["yes"] `

and there is no such key as "yes" in the object

```
let pizza = {

    name: "Peperoni",
    description: "Pizza is in normal size and with extra Mayo sauce on",
    price: "35 kr."
    ingredients = ["peperoni", "pepper", "tomato", "cheez", "onion", "mashrooms"];
}
```

PR #110, opened 2023-11-24
javascript1-week3/Daisy

Daisyangtan:


3_javascript/javascript1/week3/script2.js
```diff
@@ -0,0 +1,46 @@
...
+        hours: 12,
+        minutes: 0
+    }
+];
+
+function logOutSeriesText() {
+    const totalMinutesWatched = seriesDurations.reduce((total, series) => {
+        const seriesDurationInMinutes = series.days * 24 * 60 + series.hours * 60 + series.minutes;
+        return total + seriesDurationInMinutes;
+    }, 0);
```
BrantalikP, line 30: Amazing 🤩 especially with how you've utilized ﻿reduce. The task is definitely correct.

However, I see that you are looping through the ﻿`seriesDurations` variable twice. I believe you can streamline the logic to execute within one loop instead. Also, the calculation ﻿`series.days * 24 * 60 + series.hours * 60 + series.minutes;` appears to be repeated. You could simplify your code by storing this calculation in a separate variable. This approach not only minimizes redundancy, but also makes debugging easier in case of an error or typo in the calculation. Fixing the problem at one place would then rectify the issues throughout your code. Here's an alternative approach you might find interesting.

```
{
   let totalMinutesWatched = 0
   const averageLifespanYears = 80;


    seriesDurations.forEach(series => {
	const seriesDurationInMinutes = series.days * 24 * 60 + series.hours * 60 + series.minutes;
        totalMinutesWatched =+ seriesDurationInMinutes;

        const seriesDurationInYears = seriesDurationInMinutes / (60 * 24 * 365);
        const percentageOfLifeForSeries = (seriesDurationInYears / averageLifespanYears) * 100;
        console.log(`${series.title} took ${percentageOfLifeForSeries.toFixed(3)}% of my life`);
    });

  	const totalYearsWatched = totalMinutesWatched / (60 * 24 * 365);

  	const percentageOfLife = (totalYearsWatched / averageLifespanYears) * 100;
        console.log(`In total that is ${percentageOfLife.toFixed(3)}% of my life`);
}
```

3_javascript/javascript1/week3/script3.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Example usage
+saveNote("Pick up groceries", 1);
+saveNote("Do laundry", 2);
+console.log(notes);
+
+const firstNote = getNote(1);
+console.log(firstNote);
+
```
BrantalikP, line 51: wow, really great job, I like that you did the extra step with the delete method!

BrantalikP (approved):
As always, amazing job 👏 really like how each assignment has its own module.
Also great usage advanced methods such as `reduce` and `splice`

PR #111, opened 2023-11-25
javascript1-week3/Monika

Monika-Marciniak:
I havent done all tasks yet. Here's part I finished.

### DIFF
--- a/3_javascript/javascript1/week3/homework.html/distance.js
+++ b/3_javascript/javascript1/week3/homework.html/distance.js
@@ -1,0 +1,22 @@
+function calculateTravelTime(travelInformation) {
+    const speed = travelInformation.speed;
+    const distance = travelInformation.destinationDistance;
+  
+    const timeInHours = distance / speed;
+  
+    const hours = Math.floor(timeInHours);
+    const minutes = Math.round((timeInHours * 60) % 60);
+  
+    const timeFormatted = hours + " hours and " + minutes + " minutes";
+  
+    return timeFormatted;
+  }
+  
+  const travelInformation = {
+      speed: 50,
+      destinationDistance: 432,
+    };
+    
+    const travelTime = calculateTravelTime(travelInformation);
+  console.log(travelTime);
+  
--- a/3_javascript/javascript1/week3/homework.html/distance.js
+++ b/3_javascript/javascript1/week3/homework.html/distance.js
@@ -5,7 +5,7 @@ function calculateTravelTime(travelInformation) {
     const timeInHours = distance / speed;
   
     const hours = Math.floor(timeInHours);
-    const minutes = Math.round((timeInHours * 60) % 60);
+    const minutes = Math.floor((timeInHours * 60) % 60);
   
     const timeFormatted = hours + " hours and " + minutes + " minutes";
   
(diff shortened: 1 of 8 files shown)


3_javascript/javascript1/week3/homework.html/distance.js
```diff
@@ -0,0 +1,22 @@
+function calculateTravelTime(travelInformation) {
+    const speed = travelInformation.speed;
+    const distance = travelInformation.destinationDistance;
+  
+    const timeInHours = distance / speed;
+  
+    const hours = Math.floor(timeInHours);
+    const minutes = Math.round((timeInHours * 60) % 60);
```
SvitlanaBurlaka: I would suggest to use Math.floor as well, so you have it in both variables :)

3_javascript/javascript1/week3/homework.html/remove.js
```diff
@@ -0,0 +1,16 @@
...
+    "Ahmad",
+    "Yana",
+    "kristina",
+    "Rasmus",
+    "Samuel",
+    "katrine",
+    "Tala",
+  ];
+  const nameToRemove = "Ahmad";
+  delete names[1];
```
SvitlanaBurlaka: Interesting choice to use delete method 🤔  You can read about this here : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete . As you can see , it is better to use for object.
But for this exercise, I would recommend to use splice method : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice which is better when you have data stored in array :)

3_javascript/javascript1/week3/homework.html/tax.js
```diff
@@ -0,0 +1,22 @@
+function calculateTax(income) {
```
SvitlanaBurlaka, line 1: Is it your unique feature? I cannot find the task in homework description.
Anyway, here I would recommend to use if and else if and the the last condition will end with else, instead of multiply if's in your code. That will make your code to look more structured.

SvitlanaBurlaka (commented):
Good job! Keep it up! As I noticed, you need to finish one more exercise for seriesDuration, right?

3_javascript/javascript1/week3/homework.html/tax.js
Monika-Marciniak: It was on slides. I changed it to else if :) 
Write a function that calculates the tax percentage based on the give income
For example:
If your income is less than 50,000 you pay 8%
If your income is between 50,000 and 100,000 you pay 15%
If your income is between  100,000 and  300,000 you pay 30%
If your income is above that, you pay 50%
Try the function with different income values to verify

3_javascript/javascript1/week3/homework.html/remove.js
Monika-Marciniak: Okay done

3_javascript/javascript1/week3/homework.html/distance.js
Monika-Marciniak: done :)

PR #120, opened 2023-11-25
javascript1-week3/Viki

VictoriaDem17:
Homework for javascript1-week3.

### DIFF
--- a/3_javascript/javascript1/week3/smart-ease.js
+++ b/3_javascript/javascript1/week3/smart-ease.js
@@ -1,0 +1,65 @@
+//Javascript warmup (week3)
+
+//Exercise 1 - Item array removal
+const names = [
+    "Peter",
+    "Ahmad",
+    "Yana",
+    "kristina",
+    "Rasmus",
+    "Samuel",
+    "katrine",
+    "Tala",
+  ];
+  const nameToRemove = "Ahmad";
+  const index = names.indexOf("Ahmad");
+  if (index !==-1) {
+    names.splice(index, 1);
+  }
+  console.log(names);
+
+//Exercise 2 - When will we be there??
+const travelInformation = {
+    speed: 50,
+    destinationDistance: 432,
+  };
+
+function timeToDrive(travelInformation) {
+    let timeTotal = (travelInformation.destinationDistance / travelInformation.speed) * 60;
+    let hour = Math.floor(timeTotal / 60);
+    let minute = Math.floor(timeTotal % 60);
+    return `${hour} hours ${minute} minutes`;
+  }
+  
+  const travelTime = timeToDrive(travelInformation);
+  console.log(travelTime);
+  
+  //Series duration of my life
+  const seriesDurations = [
+    {
+      title: "Friends",
+      days: 3,
+      hours: 23,
+      minutes: 0,
+    },
+    {
+      title: "Black Books",
+      days: 4,
+      hours: 6,
+      minutes: 30,
+    },
+    {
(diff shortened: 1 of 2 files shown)


3_javascript/javascript1/week3/smart-ease.js
```diff
@@ -0,0 +1,127 @@
+//Javascript warmup (week3)
+
+//Exercise 1 - Item array removal
```
SvitlanaBurlaka, line 3: In this exercise all good 👍

3_javascript/javascript1/week3/smart-ease.js
```diff
@@ -0,0 +1,127 @@
...
+      title: "The Good Doctor",
+      days: 7,
+      hours: 18,
+      minutes: 21,
+    },
+  ];
+    const averageLifespanYears = 80;
+    const averageLifespanMinutes = averageLifespanYears * 365 * 24 * 60;
+    let allSeriesDuration = 0;
+
```
SvitlanaBurlaka, line 61: In this exercise all looks good for me 👀  But if you wish to use reduce method, you would need to use also forEach method which will be instead of for loop , not filter.
Let's say, you want to find total duration in minutes with reduce method, which will add values up. This method always(in this case) stars from 0, and then reduce method walks through element-by-element, at each step adding the current value to the result from the previous step (this result is the running sum of all the previous steps) — until there are no more elements to add. More information you can find here : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce . So the code will be like this : 
`const totalSeriesDurationInMinutes = seriesDurations.reduce((total, series) => {
  return total + (series.days * 24 * 60 + series.hours * 60 + series.minutes);
}, 0);
 `

3_javascript/javascript1/week3/smart-ease.js
```diff
@@ -0,0 +1,127 @@
...
+  saveNote("Rest a little, read a book", 5);
+
+  console.log(notes);
+
+//Get a note
+  function getNote(id) {
+   for (const note of notes) {
+    if (note.id === id) {
+    return note;
+  } else {
```
SvitlanaBurlaka, line 99: It will always throw error, because in your code it is not checking through the whole array. So basically it will stop at the first element of the array if you call `getNote(2) ` because the first note does not match given `id`. 
You have to move Error outside  of for loop like this : 
`function getNote(id) {
  for (const note of notes) {
    if (note.id === id) {
      return note;
    } else {
    }
  }
  return Error;
}`

3_javascript/javascript1/week3/smart-ease.js
```diff
@@ -0,0 +1,127 @@
...
+  };
+  logOutNotesFormatted();
+
+  //Unique feature
+/* I was thinking to connect notes array with days of week,
+when it shows what person need to do depends of day.
+For example, to feed fish and walk with dog - do every day,
+but laundry - on Wednesday and groceries - on Monday. I would appreciate if
+you (mentor who's checking my work) could give me some tips how to create it, and
+where to start. */
```
SvitlanaBurlaka, line 127: I love the idea! What I can think about here is to modify your `const notes` so you will have an array of objects, where instead of id: number will be day of the week :
```
const tasksByDay = [
  { id: "Monday", tasks: ["Groceries"] },
  { id: "Tuesday", tasks: ["Feed Fish", "Walk with Dog"] },
  { id: "Wednesday", tasks: ["Laundry"] }, and so on
```

PR #122, opened 2023-11-25
javascript1-week3/sanaz

Sanazre:


3_javascript/javascript1/week3/app.js
```diff
@@ -0,0 +1,172 @@
...
+  "Tala",
+];
+const nameToRemove = "Ahmad";
+
+const filteredNames = names.filter((name) => name !== nameToRemove);
+
+console.log(filteredNames);
+
+//2 -----------------------------
+// Write a function where you speficy your speed in km/h and how far you have to go in km.
```
BrantalikP, line 20: Hey! I noticed there is a lot of commented code. If we aren't using it, could you please delete it? The code should be as clean as possible in case you need to work in a team. These unused peaces can be distracting. Thank you!

BrantalikP (changes requested):
Good job so far @Sanazre 👏 but please also include [Series duration of my life ](https://github.com/HackYourFuture-CPH/JavaScript/blob/main/javascript1/week3/homework.md#series-duration-of-my-life) assignemnt

BrantalikP (approved):
Well done 👏

PR #123, opened 2023-11-25
javascript-week3/yousseftopaji

yousseftopaji:


3_javascript/javascript1/week3/js.js
```diff
@@ -0,0 +1,138 @@
+// Item array removal
+
```
BrantalikP, line 2: `js.js` its sounds cryptic :D Try to stay with connections and name it maybe `script.js` or `week3.js` so it's clear what the file is doing next time. Thanks!

3_javascript/javascript1/week3/js.js
```diff
@@ -0,0 +1,138 @@
...
+    "Yana",
+    "kristina",
+    "Rasmus",
+    "Samuel",
+    "katrine",
+    "Tala",
+  ];
+  const nameToRemove = "Ahmad";
+  
+  const secondName = names.splice(1,1);
```
BrantalikP, line 17: Haha, that works! :D However, it only runs properly while Ahmad resides at index 1. The function needs to be more dynamic. Please check this [out](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)

3_javascript/javascript1/week3/js.js
```diff
@@ -0,0 +1,138 @@
...
+
+
+//   When will we be there??
+
+
+
+function timeCounter(speed, destinationDistance){
+    let counter= destinationDistance/speed;
+    let hour = Math.floor(counter);
+    let minutes = Math.floor ((counter-hour)*60);
```
BrantalikP, line 33: for this one, I would rather use` Math.round ` What do you think? In my opinion the number should decide by itself whether it wants to round up or down, but it's not an issue. Good job!

3_javascript/javascript1/week3/js.js
```diff
@@ -0,0 +1,138 @@
...
+function yourAge(age) {
+    
+let result="";
+let resultInTotal=0;
+    for (const seriesDuration of seriesDurations){
+        let ageInMinutes = age*365*24*60;
+        let seriesDurationsInMinutes = (seriesDuration.days*24*60) + (seriesDuration.hours*60) + seriesDuration.minutes;
+        let wastedLifePercentage = (seriesDurationsInMinutes/ageInMinutes)*100;
+
+        result = `${seriesDuration.title} has wasted ${wastedLifePercentage.toFixed(3)} % of your life`;
```
BrantalikP, line 75: Since the `result` variable is used only within this loop. you can declare it here like 

`const result = `${seriesDuration.title} has wasted ${wastedLifePercentage.toFixed(3)} % of your life`;`

3_javascript/javascript1/week3/js.js
```diff
@@ -0,0 +1,138 @@
...
+        console.log (result);
+        resultInTotal+=wastedLifePercentage;
+    }
+    console.log(`in total that is ${resultInTotal.toFixed(3)} of your life!`);
+
+}
+
+yourAge (70);
```
BrantalikP, line 83: nice! Good job! I Like that you made this function dynamic! 👏

3_javascript/javascript1/week3/js.js
```diff
@@ -0,0 +1,138 @@
...
+saveNote("Do laundry", 2);
+
+console.log(notes);
+
+
+// GetNote
+
+
+function getNote(id){
+    for (const note of notes){
```
BrantalikP, line 108: works well. please implement this part of the assignment also` If no id is specified or if the id is not a number, log out an error string.`

BrantalikP (commented):
Well done @yousseftopaji 💪 HW is done sir! only please add the check for `id` in the `notes` assignment. Thanks :)

PR #128, opened 2023-11-28
javascript1-week3/anastasiia

humenyuk16:


3_javascript/javascript1/week3/script.js
```diff
@@ -0,0 +1,136 @@
...
+    "kristina",
+    "Rasmus",
+    "Samuel",
+    "katrine",
+    "Tala",
+  ];
+  const nameToRemove = "Ahmad";
+  let index = names.indexOf("Ahmad");
+  console.log(index); //index "Ahmad" in the array 1
+  if (index > 0){
```
SvitlanaBurlaka, line 17: i would check here like this : `if(index>=0)`, because the first element in the array has 0 index, and maybe you would want to remove it in future 😸

SvitlanaBurlaka (commented):
You have also files here hor homework two, so you can delete them and have only for week three, since you have separate PR for that homework. Your code looks good! Keep it up ⭐

PR #129, opened 2023-11-29
javascript1-week4/hadis

Hadis-jamali:


3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,77 @@
...
+
+function getReply(command) {
+  if (command.startsWith("Hello my name is")) {
+    personName = command.split(" ")[4];
+    console.log(`nice to meet you ${personName}`);
+    return;
+  }
+
+  if (command === "What is my name") {
+    if (personName === "") console.log("I dont have your name");
```
NVedsted, line 13: Omitting the curly braces for this `if`-statement is valid, but it can be tricky to work with; especially if you come back to it and need to make changes! I'd suggest consistently sticking to always wrapping the `if`-blocks inside curly braces. 😄

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  if (command === "What day is it today?") {
+    let date = new Date();
+    console.log(`Today is ${date}`);
+  }
+  if (command.startsWith("What is")) {
+    let math = "";
+    math = command.substring(8, command.length);
+    console.log(eval(math));
```
NVedsted, line 57: This is a clever choice that allows for quite complex math as well! However, it is highly dangerous to use this function, and I will recommend avoiding. As an example, I could ask `What is alert('You have been hacked!')` and it would show a message on the screen. This could be used by hackers to stir up trouble! 😱

I'd suggest simplifying the math you will allow and code something simple that handles the most basic operations on two numbers such that you can avoid calling `eval`.

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,77 @@
...
+    //   if (i == commandArr.length - 4)
+    // toDo += commandArr[i];
+    //   else toDo += commandArr[i] + " "; }
+
+    let toDo = "";
+    toDo = command.substring(4, command.length - 11);
+    toDoList.push(toDo);
+    console.log(`${toDo} added to your todo`);
+  }
+  if (command.startsWith("Remove")) {
```
NVedsted, line 33: This `if` and others below will be checked even if the above case was hit. To make it easier to manage and less error prone, I'd suggest using `else if` syntax. The immediate benefit is that you do not need to return when a command has been processed and you do risk running multiple handlers for the same command; it is also marginally faster! 🚀

NVedsted (changes requested):
Nice solution to the assignment! 🌟 I have left a few comments for you; mainly to avoid `eval` as it can be quite dangerous. Let me know if you encounter any issues! Have a nice evening. 🙂

PR #131, opened 2023-11-30
javascript1-week4/Niloufar

Niloufar97:


3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,122 @@
+let name = "";
+let todos = [];
+const getReply = (command) => {
+  let splittedCommand = command.split(" ");
+  // name ----------------------------------------------------------
+  if (command.includes("my name is")) {
+    let nameIndex = splittedCommand.indexOf("name");
```
MJanOthman, line 7: I liked the way you are detecting the index of the name, great one.

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,122 @@
...
+    if (nameIndex + 2 < splittedCommand.length) {
+      if (name === "") {
```
MJanOthman, line 9: you should add also `&&  splittedCommand[nameIndex + 2]!== " "`
I tried to add 2 spaces between 'is' and 'Niloufar', it was taking the space as the name, so I got 
`nice to meet you `

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,122 @@
...
+
+//   date ------------------------------------------
+    if(command.includes("what day is today")){
+        const today = new Date();
+        const options = {day: 'numeric' , month: 'long' , year: 'numeric'};
+        const formatedDate = today.toLocaleDateString('en-US', options)
+        return `today is ${formatedDate}`
+    }
+// math--------------------------------------------------------------
+
```
MJanOthman, line 68: This works exactly as wanted, well done.

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,122 @@
...
+    if (command.includes("set timer for")) {
+        const forIndex = splittedCommand.indexOf('for');
+        const time = parseInt(splittedCommand[forIndex + 1]);
+        setTimeout(() => {
+            console.log('Timer done');
+        }, time * 1000 * 60 ); 
+    
+        return(`Timer set for ${time} minutes`) ;  
+    }
+};
```
MJanOthman, line 103: To stay at safe side, you can return a string here, that will show at any case that is not defined by you inside the function, fx:
now if you use any wrong command, it will return `undefined`
so I am suggesting something like this:

```suggestion
 return "I don't understand you";
```

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,122 @@
+let name = "";
+let todos = [];
+const getReply = (command) => {
+  let splittedCommand = command.split(" ");
+  // name ----------------------------------------------------------
```
MJanOthman, line 5: nit:
What if the user uses Upper case for the first word in the sentence?
fx: `What is my name?`
to make the code safe, I recommend using `.tolowerCase()`, so it will not be case sensitive anymore.

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,122 @@
...
+        const forIndex = splittedCommand.indexOf('for');
+        const time = parseInt(splittedCommand[forIndex + 1]);
+        setTimeout(() => {
+            console.log('Timer done');
+        }, time * 1000 * 60 ); 
+    
+        return(`Timer set for ${time} minutes`) ;  
+    }
+};
+// name--------------------------------------------------------------
```
MJanOthman, line 104: Clean, readable, and understandable code, I like the way you return inside the function and then console.log when you are calling the function, this is the right approach.

MJanOthman (commented):
Really great job, the functionality is cool.
I have added some small comment to make the code a bit more safe.

PR #133, opened 2023-12-01
javascript1-week4/Monika

Monika-Marciniak:


3_javascript/javascript1/week4/voiceassistant.js
```diff
@@ -0,0 +1,92 @@
+
+let name = '';
+let todos = [];
+let timer;
+
+function getReply(command) {
+  if (command.startsWith('Hello my name is')) {
+    name = command.replace('Hello my name is', '');
```
BrantalikP, line 8: nice, this is smart! another approach could be to split the string into words that could be accessed via index e.g 
```
splittedCommand = command.split(' '). 
splittedCommand[4] // Monika
```

BrantalikP (approved):
Great Job! the code is well designed and clean. Love it 👏

PR #137, opened 2023-12-02
javascript1-week4/maryam

MaryamAarhus:


3_javascript/javascript1/week4/homework4.js
```diff
@@ -0,0 +1,156 @@
...
+
+console.log(processData(mockData));
+
+//Voice assistant
+
+let userName = "";
+let todos = [];
+
+function getReply(command) {
+    const commandLower = command.toLowerCase();
```
SvitlanaBurlaka, line 93: good idea to use toLowerCase 👍

SvitlanaBurlaka (commented):
Good, structured and easy to read! The one question, is there suppose to be folders, files for previous homeworks? If not, then clean it and have only for week4 :)

PR #138, opened 2023-12-02
javascript1-week4/Gizem

GizemSavci:


SvitlanaBurlaka (commented):
Good job! Have nothing to add, keep it uo ⭐

3_javascript/javascript1/week4/index.js
```diff
@@ -0,0 +1,59 @@
+let name = "";
```
SvitlanaBurlaka, line 1: one more thing, I would rename this variable, because ' name' is deprecated. Or you can just move the variable inside the func getReply

PR #139, opened 2023-12-02
javascript1-week4/anjali

anjali744:
There is something wrong in my code i couldn't able to find that error, In this part "Simple Arithmetic Calculation". Thankyou in advance. 🙏🏻 ☺️

### DIFF
--- a/3_javascript/javascript1/week4/index.html
+++ b/3_javascript/javascript1/week4/index.html
@@ -19,18 +19,18 @@
     <div class="arithmeticContainer">
       <h2>Simple Arithmetic Calculation</h2>
       <hr />
-      <input type="number" placeholder="Number1" />
+      <input type="number" id="num1" placeholder="Number1" />
       <br />
-      <input type="number" placeholder="Number2" />
+      <input type="number" id="num2" placeholder="Number2" />
       <br />
       <div class="button-row">
-        <button>+</button>
-        <button>-</button>
-        <button>/</button>
-        <button>*</button>
+        <button id="add">+</button>
+        <button id="sub">-</button>
+        <button id="div">/</button>
+        <button id="mul">*</button>
       </div>
       <br />
-      <input type="number" readonly />
+      <input type="number" id="result" readonly />
     </div>
     <script src="script.js"></script>
   </body>
--- a/3_javascript/javascript1/week4/index.html
+++ b/3_javascript/javascript1/week4/index.html
@@ -1,0 +1,37 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>js1-week4</title>
+    <link rel="stylesheet" href="styles.css" />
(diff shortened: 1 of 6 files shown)


3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,113 @@
...
+  return "Timer set for 4 minutes.";
+}
+
+console.log(getReply("Hello my name is Benjamin"));
+console.log(getReply("What is my name?"));
+console.log(getReply("Add fishing to my todo"));
+console.log(setTimer(4));
+console.log(getReply("Open website www.example.com"));
+
+//Simple Arithmetic Calculation-----------  //there is somthing wrong in this code it does't work.
```
SvitlanaBurlaka, line 95: In your HTML file, the input elements for numbers and the result field are missing the id attribute. Without the id, your JavaScript code won't be able to recognise which element you are trying to get.

3_javascript/javascript1/week4/index.html
```diff
@@ -0,0 +1,37 @@
...
+        This is my codeWars link is here.
+        <a href="https://www.codewars.com/users/anjali744"
+          >Visit my CodeWars profile</a
+        >
+      </p>
+    </div>
+    <div class="arithmeticContainer">
+      <h2>Simple Arithmetic Calculation</h2>
+      <hr />
+      <input type="number" placeholder="Number1" />
```
SvitlanaBurlaka: You need to update it like this : 
```
<input type="number" id="num1" placeholder="Number1" />
<br />
<input type="number" id="num2" placeholder="Number2" />
<br />
<div class="button-row">
  <button id="add">+</button>
  <button id="sub">-</button>
  <button id="div">/</button>
  <button id="mul">*</button>
</div>
<br />
<input type="number" id="result" readonly />
```

3_javascript/javascript1/week4/script.js
```diff
@@ -0,0 +1,113 @@
...
+    console.log("Timer done");
+  }, 4 * 60 * 1000);
+
+  return "Timer set for 4 minutes.";
+}
+
+console.log(getReply("Hello my name is Benjamin"));
+console.log(getReply("What is my name?"));
+console.log(getReply("Add fishing to my todo"));
+console.log(setTimer(4));
```
SvitlanaBurlaka, line 92: in this specific case, you do not need to call func as `setTimer(4)`, because you specified already 4 minutes. BUT if you want to make function more flixible and reusable, you can pass to that function as argument 'minutes' and specify them instead of hardcoded value. I will be like this: 
` function setTimer(minutes) {
  clearTimeout(timer);

  timer = setTimeout(() => {
    console.log("Timer done");
  }, minutes * 60 * 1000);

  return `Timer set for ${minutes} minutes.`;
}


console.log(setTimer(5));
`

SvitlanaBurlaka (commented):
Hi @anjali744 ! Looks good, left some minor comments! Also, why do you have in this PR files for homework3? Try to keep it organised , and leave only files for hm4 ⭐

PR #140, opened 2023-12-02
ver1

Sanazre:


3_javascript/javascript1/week4/VoiceAssistant.js
```diff
@@ -0,0 +1,79 @@
...
+// Set a timer for 4 minutes - Should respond with "Timer set for 4 minutes". When 4 minutes is up: "Timer done". How do we set a timer in js? Google is your friend here!
+// Add one or more command to your voice assistant
+
+
+
+
+let userName = ""; // متغیر جهت ذخیره نام کاربر
+let todos = []; // لیست وظایف کاربر
+function getReply(command) {
+    const lowerCommand = command.toLowerCase(); // تبدیل دستور به حروف کوچک
```
SvitlanaBurlaka, line 27: like how you are using toLowerCase , making it not sensitive ⭐

SvitlanaBurlaka (commented):
Keep up the good work, code is clear and readable 👍

3_javascript/javascript1/week4/VoiceAssistant.js
Sanazre: I appreciate your mention  do you mean that using toLowerCase allows the comparison to be case-insensitive? So, whether someone types in uppercase or lowercase, it won't affect the understanding or recognition of the input? If there's anything else you're want to mention about, feel free to let me know

SvitlanaBurlaka: I meant that it is good practice to use it, you are making it in this way **not** sensitive, and if someone will decide to ask 'WhaT iS mY nAmE?' it will make the string look organised and your 'data' will be in the same format, in your case - in lower case 👍

PR #151, opened 2024-01-30
databases-week1/anastasiia

humenyuk16:


homework.sql
```diff
@@ -0,0 +1,59 @@
+-- Active: 1706440038860@@127.0.0.1@3307@hyf_lesson1
+-- Find out how many tasks are in the task table
+SELECT count(*) FROM task;
```
veronicaburduzhan, line 3: All SQL commands should be written in uppercase 😊  
`SELECT COUNT(*) FROM task`

homework.sql
```diff
@@ -0,0 +1,59 @@
...
+
+-- Find out how many tasks in the task table do not have a valid due date
+
+SELECT count(*) FROM task 
+WHERE due_date <= NOW();
```
veronicaburduzhan, line 8: Technically this is correct but I assume the task was to check for nullish or invalid dates. Because if your due dates expired they are still valid existing dates 😊

homework.sql
```diff
@@ -0,0 +1,59 @@
...
+
+-- Find all the tasks that are marked as done
+
+SELECT * FROM task
+WHERE status_id = '3';
```
veronicaburduzhan, line 13: I left a huge comment in the slack channel explaining why this is slightly wrong 😅

homework.sql
```diff
@@ -0,0 +1,59 @@
...
+
+-- Find all the tasks that are not marked as done
+
+SELECT * FROM task
+WHERE status_id != '3';
```
veronicaburduzhan, line 18: This is same as previous :blush:

veronicaburduzhan (changes requested):
Good job  👏  Looks like you grasped all the SQL basics from the start! I left a few comments for improving your code and working with SQL in general.

PR #162, opened 2024-02-02
databases-week1/maryam

MaryamAarhus:


5_databases/week1/homework1.sql
```diff
@@ -0,0 +1,69 @@
...
+    t.id AS TaskID,
+    t.title AS Title,
+    t.description AS Description,
+    t.created AS CreatedDate,
+    t.due_date AS DueDate,
+    s.name AS Status
+FROM task AS t
+    JOIN status AS s ON t.status_id = s.id
+WHERE
+    s.name = 'Done';
```
veronicaburduzhan, line 28: However this is a valid syntax (s.name, t.id, etc), I don't this that you should use it in a real situation. Imagine you have 10-15 different tables, and it's just hard to remember each of them. The names are already short and simple but they give you an understanding of the content inside but just a letter - don't. You can leave you code as it is but keep in mind my advice for the future😉

veronicaburduzhan (approved):
Excellent job✨ You definitely understood basic concepts of SQL, keep going!

PR #169, opened 2024-02-06
databases-week2/Niloufar

Niloufar97:


5_databases/week2/part2.sql
```diff
@@ -0,0 +1,49 @@
...
+
+--Create an index on the name column of the student table.
+CREATE INDEX student_name_index ON students (student_name);
+
+--Add a new column to the class table named status which can only have the following values
+ALTER TABLE class
+ADD COLUMN class_status ENUM ("not-started", "ongoing", "finished");
+
+-- practice more
+INSERT INTO
```
NVedsted, line 33: Good job! 🎉 Love that the data reflects the HYF class moduls! 😄

5_databases/week2/part3.sql
NVedsted, line 1: These `SELECT`s and `JOIN`s are spot on, but I think you have used the wrong tables and data. 😢 The [week 2 dataset](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson2/lesson2-data.sql) should be used here which uses a many-to-many relation through a table called `user_task` instead of `user_id` directly on `task`.

Let me know if this causes you issues. 😄

5_databases/week2/ERD.pdf
NVedsted, line 1: Very nice and concise modelling domain you have chosen. 😄 I am not entirely sure this qualifies as an ER diagram; I would perhaps take a look at [this diagram](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson2/hyf_db_mm_eer_diagram.png) for inspiration. With that being said, I think I understand what you have modelled, and I will try to add some remarks based on this understanding:
 - The borrows relation seems to be a one-to-many relation which would imply that a member can borrow many books but a book can only be borrowed by one person. This is very much true at a specific point in time,  but since the relation also stores historic data , i.e. a borrow that has a return date, this should probably be a many-to-many relation.
 - This is very nit-picky, but the is-located-in relation between a book and a rack might be unsound if the book is borrowed by someone. In that case the, it should probably be possible for it to not be located in any rack. Or perhaps the relation rather signifies something like belongs-in a certain rack.
 - It would also be nice with some types on the columns of the tables as well as showing primary keys and foreign keys.

Please correct me if I have misunderstood your modelling! I'd suggest redoing it in a format closer to the example linked above. 😄

5_databases/week2/part4.sql
NVedsted, line 1: This seems to accurately reflect your diagram according to my understanding and also seems to skirt around all the issues I mentioned with the ER diagram. Good job! 😄 I had trouble running the creation script as `racks` was referenced in `books` before it was defined, but other than that this model is in tip-top shape! Awesome! 👍

NVedsted (changes requested):
Awesome work! ⭐ You seem to have a solid grasp on SQL! 😄 I have left the majority of comments on part 4 which I'd recommend spending the most time on.

Happy coding! 🎉

PR #171, opened 2024-02-06
databases-week2/angel

AngelHenriettaAboah:
Hey there... here's a pull to my databases week 2 homework and exercises. Happy Coding!!


5_databases/week2/hwkpart2data.sql
```diff
@@ -0,0 +1,42 @@
+-- Active: 1706445451493@@127.0.0.1@3306@School_database
+CREATE TABLE school_class (
+    id INT PRIMARY KEY AUTO_INCREMENT,
+    name VARCHAR(255) NOT NULL,
+    start_date DATE,
+    end_date DATE,
+    status ENUM('not-started', 'on-going', 'finished') DEFAULT 'not-started'
```
NVedsted, line 7: Great work on these `CREATE`-statements! 🎉 I think the homework intended for you to add the `status` column through an `ALTER`-statement. How would that look?

5_databases/week2/hwkpart3task.sql
NVedsted, line 1: Nice work with the queries! And very interesting to see the use of sub-queries! 😮 I think you have used the wrong tables and data. 😢 The [week 2 dataset](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson2/lesson2-data.sql) should be used here which uses a many-to-many relation through a table called `user_task` instead of `user_id` directly on `task`.

5_databases/week2/hwkpart3task.sql
```diff
@@ -0,0 +1,29 @@
...
+
+SELECT * FROM task
+JOIN user ON task.user_id = user.id
+WHERE user.email LIKE '%@spotify.com';
+
+Get all the tasks for 'Donald Duck' with status 'Not started'
+
+SELECT * 
+FROM task
+WHERE user_id = (SELECT id FROM user WHERE name = 'Donald Duck')
```
NVedsted, line 13: You do not have to change anything, but I want you to briefly consider how you would do this with joins instead? 😄

5_databases/week2/Blank erdiagramweek2 - Page 3.png
NVedsted, line 1: I assume this is the ER diagram for the homework? Correct me if I am wrong. 😄 I'll put my feedback for the entire diagram in bullet form:
 - Firstly, the relations are hard to decipher as they are all connected to each other. In ER diagrams, a relation should only have two endpoints, E.g. in this diagram, it would be clearer if the connections were directly between the tables.
 - I am not sure the appropriate symbols are being used for the relations. You could take a look at [this diagram](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson2/hyf_db_mm_eer_diagram.png) for inspiration. It would also be nice to have some types on the different columns, e.g. that age is an integer, name a string, etc.
 - How does the farmer fit into the model? Perhaps a many-to-one relation to `Barn`?
 - The model seems to imply, through the `Barn_ID` FK on `Animal` that an animal can only belong to a single barn, however, the `Barn_Animal` table seems to be a many-to-many relation between animals and barns, i.e. that an animal can belong to multiple barns and vice versa. With the being said, very nice that you are showing the linking table for the many-to-many relation!

Let me know if anything causes issues with this. 😄

5_databases/week2/hwkpart4data.sql
NVedsted, line 1: This seems to correctly reflect your ER diagram; very good work! 🎉 And nice with the `INSERT`s to have some sample data to look at. 😄

NVedsted (changes requested):
Great job on the homework! ⭐ I have left the majority of comments on the ER diagram and I'd suggest spending most of your time there since you seem to have a solid grasp on the rest. 💪 

Happy coding!

5_databases/week2/hwkpart2data.sql
AngelHenriettaAboah: > Great work on these `CREATE`-statements! 🎉 I think the homework intended for you to add the `status` column through an `ALTER`-statement. How would that look?

yes obviously but I was just being a lazy programmer...:

PR #172, opened 2024-02-08
databases-week2/katayoun

katayouny:


5_databases/week2/part1-working-with-tasks.sql
```diff
@@ -0,0 +1,37 @@
...
+
+
+-- Add a task with these attributes: title, description, created, updated, due_date, status_id, user_id
+INSERT INTO task (title, description, created, updated, due_date, status_id, user_id) VALUES ('doing homeworks', 'SQL homeworks week2', '2018-09-26 03:06:46', '2018-10-08 06:14:31', '2019-12-22 20:58:03', 3, 8);
+
+
+-- Change the title of a task
+UPDATE task
+SET title='doing excercises'
+WHERE user_id=6;
```
NVedsted, line 13: This updates all the tasks made by the user with ID 6 and not just one task, i.e. every single task for user 6 will get the title `doing exercises`. I am not sure if this is intentional. 😄

The same comment applies to all the other `UPDATE`-statements in part 1.

5_databases/week2/part1-working-with-tasks.sql
```diff
@@ -0,0 +1,37 @@
...
+
+-- Change a task due date
+UPDATE task
+SET due_date='2020-09-26 03:06:46'
+WHERE user_id=6;
+
+
+-- Change a task status
+UPDATE task
+SET status_id='2'
```
NVedsted, line 24: It is a bit misleading to type `'2'` instead of `2` here as it is a number and not a string. MySQL is of course kind enough to translate for us. 😄

5_databases/week2/part3-more-queries-hyf-database.sql
```diff
@@ -0,0 +1,32 @@
+use hyf;
+
+select * from user;
+
+--Get all the tasks assigned to users whose email ends in @spotify.com
+SELECT task.title as 'task title'
```
NVedsted, line 6: These `SELECT`s and `JOIN`s are spot on, but I think you have used the wrong tables and data. 😢 The [week 2 dataset](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson2/lesson2-data.sql) should be used here which uses a many-to-many relation through a table called `user_task` instead of `user_id` directly on `task`.

Let me know if this causes you issues. 😄

5_databases/week2/part3-more-queries-hyf-database.sql
```diff
@@ -0,0 +1,32 @@
...
+JOIN status on task.status_id=status.id
+WHERE user.name='Donald Duck' AND status.name='Not started';
+
+
+--Get all the tasks for 'Maryrose Meadows' that were created in september
+SELECT task.title
+FROM task
+JOIN user ON task.user_id = user.id
+WHERE user.name = 'Maryrose Meadows'
+  AND MONTH(task.created) = 9;
```
NVedsted, line 25: Great use of the built-in function `MONTH`! 🎉

5_databases/week2/Liberary ER diagram.png
NVedsted, line 1: Neat concepts you've decided to model! 😄 I am a bit unsure what the relation between `book` and `member` means. If it's what members have borrowed a book, it would likely be easier to integrate this into `borrowing_log` and do the many-to-many relation through that.

I'll just quickly summarize some observations:
 - `genre_id` is the PK but `book` uses `genre_name` for its FK. It might be easier to stick to using one of them. Which one would you prefer?
 - Very nice to see the many-to-one relation between `borrowing_log` and `book`! It does seem like the relation has been flipped based on the columns though as in this case there can only be one log on a book. The `borrowing_log` should probably keep an ID to `book` such that multiple logs can reference the same book and a log must always be associated with one, and only one, book.
 - For the many-to-many relation between members and books, there is also an issue as the book only allows a singular reference. To model the many-to-many relation, you will need a linking table similar to `user_task` in [the week 2 SQL](https://github.com/HackYourFuture-CPH/databases/blob/main/lesson2/lesson2-data.sql).
 - The relation between racks and genres seems to have the appropriate columns, but the relation seems off. A genre can only be associated with a single rack and vice versa? This might be what you intended, but I will ask if it should be possible to have multiple racks for the same genre or perhaps a mixed-genre rack? I'll leave that up to you. 😄

This was a bit of a mothful; let me know if anything seems off. 🙂

5_databases/week2/creating-a-database-liberaryDB.SQL
NVedsted, line 1: This seems to accurately represent your diagram; good job! 🎉 There was a few issues for me running your script such as `genre` being referenced in `rack` before the former was defined. MySQL also wasn't happy about `genre` itself as the auto column is not the key.

Nice touch with the index on book title! That will make it a lot more efficient to search for books. 😃

NVedsted (changes requested):
Good job with the homework! ⭐ There are some issues, but generally it is very high quality! I put a few more comments on here than I probably should, so feel free to pick and choose between what you want to focus on. The final part with the ER diargram and creating the database is definitely of the most important, so I would focus on those!

Happy coding! 😄

katayouny:
Hey Nicklas @NVedsted 

Thank you so much for your review and suggestions :)

PR #174, opened 2024-02-09
databases-week2/anastasiia

humenyuk16:


5_databases/week2/create-db-template.sql
```diff
@@ -0,0 +1,16 @@
+-- Active: 1706440038860@@127.0.0.1@3307
+CREATE DATABASE school_db DEFAULT CHARACTER SET = 'utf8mb4';
+
+USE school_db;
+
+CREATE TABLE class (
```
MahmoudYounes, line 6: this is a nitty picky comment so it's up to you to address. while your queries are correct, it is a convention to define every column on a separate line. it is cleaner and more readable. in real world, this convention is a must.

5_databases/week2/homework_2 part 3.sql
```diff
@@ -0,0 +1,36 @@
+-- Get all the tasks assigned to users whose email ends in @spotify.com
+use hyf_lesson1;
+
+SELECT *
+FROM task
```
MahmoudYounes, line 5: task table has no user_id. you should start joining from user_task table. this should be how the database looks like no? https://github.com/HackYourFuture-CPH/databases/blob/main/lesson2/lesson2-data.sql

5_databases/week2/homework_2_part1.sql
```diff
@@ -0,0 +1,27 @@
+-- Active: 1706440038860@@127.0.0.1@3307@hyf_lesson1
+-- Add a task with these attributes: title, description, created, updated, due_date, status_id, user_id
+-- Change the title of a task
+-- Change a task due date
+-- Change a task status
+-- Mark a task as complete
+-- Delete a task
+
+INSERT INTO
```
MahmoudYounes, line 9: you are missing an ERD diagram. so the comment you placed on the diff does not show the ERD. you can just add the ERD file to the repo like any other file, commit, and push.

5_databases/week2/create-db-template.sql
humenyuk16: @MahmoudYounes  yes, I totally agree with that:) But, when I save my file in vscode it is automatically  formatting all in this way like you see

5_databases/week2/homework_2 part 3.sql
humenyuk16: @MahmoudYounes  it is must be, like this?

SELECT 
task.id,
task.title,
task.user_id,
user.name,
user.email
FROM task
    JOIN user ON user.id = task.user_id
WHERE
    user.email LIKE '%@spotify.com';

PR #177, opened 2024-02-10
databases-week2/Helena

helena69hme:


5_databases/week2/part4.sql
```diff
@@ -0,0 +1,43 @@
+USE farm;
```
MahmoudYounes, line 1: This implementation does not reflect your ERD diagram :sweat_smile: 
the issue is your many to many relationship are not represented correctly.
farmer and dairyproducts is a many to may relationship which means you are going to need a farm_dairyproduct table. so the rule is one to one, one to many and many to one can be represented in the same table. many to many implies a third table to hold the relation.
This applies also to all the many to many relationship you defined in the ERD Diagram

5_databases/week2/part3.sql
```diff
@@ -0,0 +1,23 @@
+USE hyf_mysql
+
+--Get all the tasks assigned to users whose email ends in @spotify.com
+SELECT * FROM task;
+
+SELECT * FROM task JOIN user ON task.user_id = user.id WHERE user.email LIKE '@spotify.com%' ;
```
MahmoudYounes, line 6: task does not have a user_id column so you need to start from user_task table.

5_databases/week2/part3.sql
```diff
@@ -0,0 +1,23 @@
...
+
+--Get all the tasks for 'Donald Duck' with status 'Not started'
+SELECT * FROM task
+JOIN user ON task.user_id = user.id
```
MahmoudYounes, line 10: same as above comment

5_databases/week2/part3.sql
```diff
@@ -0,0 +1,23 @@
...
+JOIN status ON task.status_id = status.id
+WHERE user.name = 'Donald Duck' AND status.name = 'Not started';
+
+--Get all the tasks for 'Maryrose Meadows' that were created in september (hint: month(created)=month_number)
+SELECT * FROM task
+JOIN user ON task.user_id = user.id
```
MahmoudYounes, line 16: same as above comment

5_databases/week2/part2.sql
```diff
@@ -0,0 +1,26 @@
...
+    ends DATE NOT NULL
+)
+
+--create a 'student' table with the columns: id, name, email, phone, class_id (foreign key)
+CREATE TABLE student(
+    student_id INT PRIMARY KEY AUTO_INCREMENT,
+    student_name VARCHAR(50) NOT NULL,
+    email  VARCHAR(50) NOT NULL,
+    phone VARCHAR(50) NOT NULL,
+    class_id INT
```
MahmoudYounes, line 18: this is not defined as a foreign key :grin: need to fix that

PR #181, opened 2024-02-12
databases-week3/hadis

Hadis-jamali:


5_databases/week3/homework3.sql
```diff
@@ -0,0 +1,85 @@
...
+
+CREATE TABLE review (
+    id INT(6),
+    title VARCHAR(20),
+    description TEXT,
+    meal_id INT(6),
+    FOREIGN KEY (meal_id) REFERENCES meal(id),
+    stars INT,
+    create_date DATE
+);
```
veronicaburduzhan: More like a suggestion: 
1) it's better to assign foreign keys at the end. you can have a few of them and it's just more readable 
2) you could also add `NOT NULL` for columns that should be filled, fx. id should always be defined

5_databases/week3/homework3.sql
```diff
@@ -0,0 +1,85 @@
...
+    meal_id INT(6),
+    FOREIGN KEY (meal_id) REFERENCES meal(id),
+    created_date DATE,
+    contact_phonenumber VARCHAR(20),
+    contact_name VARCHAR(50),         
+    contact_email VARCHAR(100)        
+);
+
+CREATE TABLE review (
+    id INT(6),
```
veronicaburduzhan: You forgot to add `AUTO_INCREMENT PRIMARY KEY` to you id in this table 😊

veronicaburduzhan (approved):
Look good 😊  I left 2 comments to make your code better. You could also add more data to your table to play around more.

PR #184, opened 2024-02-15
databases-week2/olesia

AlesyaMazurenko:


5_databases/week2/Query_office.sql
```diff
@@ -0,0 +1,48 @@
+-- Active: 1706443440629@@127.0.0.1@3306@officce
+
```
veronicaburduzhan, line 2: It's not a mistake but more like a piece of advice for good code practices: 
each table has an `id` as a primary key, it's obviously unique, and we always call this column `id`. Foreign keys typically have a name like `status_id` that indicates that this is an id from the status table.  
Your stuff table has a PK `Person_ID` but check-ins table has a FK key `Person_id`. And this is already a bit confusing 😊  Imagine database with 10-15 tables.

veronicaburduzhan (approved):
Very good ✨  Left 1 comment to make your code even better

PR #188, opened 2024-02-17
databases-week3/anjali

anjali744:


MercedesUbeira (approved):
Hey Anjali! Really nice work here! ❤️ 

Some small comments: 

•You forgot to checkout from main for this branch, and that is why I see this homework mixed with other ones. 

•A suggestion for improving this code would be to have NOT NULL for some of the values on the tables, to make sure that they can't be empty.

•Also, perhaps it would be nice to have DATETIME instead of DATE? so you also get the time.

anjali744:
Hi Mechii!

Thank you for reviewing my homework. I corrected all the errors you pointed out, and I hope it's now up to standard. If you find any remaining issues, please let me know so I can fix them promptly.

I apologize for forgetting to checkout from the main branch. I have corrected that oversight. Additionally, I've taken care of the issue with the node modules.

Thank you for your patience and guidance.

PR #190, opened 2024-02-19
node-week1/niloufar

Niloufar97:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,55 @@
+//  Parse Arguments using process.argv
+let avg;
+function average() {
+  let sum = 0;
+  let count = 0;      
+  if(process.argv.length < 5){
```
LucasIversen, line 6: Good check! You could set the limit at 4, since we can also find the average between two numbers, but that's not really important 👌

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,55 @@
...
+        console.log("please enter at least three numbers");
+        return
+  }
+  for (let j = 2 ; j < process.argv.length; j++) {
```
LucasIversen, line 10: All good 👍

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,55 @@
...
+  avg = sum / count;
+  console.log(`avg: ${avg}`);
+}
+average()
+
+
+//  Passing Arguments Using Yargs
+import yargs from 'yargs'
+import { hideBin } from 'yargs/helpers'
+const argv = yargs(hideBin(process.argv))
```
LucasIversen, line 28: Good job 👍  If you want to make the yargs implementation more flexible, you can also allow users to input an array of numbers, making it able to handle more than 3 numbers

LucasIversen (approved):
Really good job trying out both ways of handling input from the console!

PR #192, opened 2024-02-19
node-week1/hadis

Hadis-jamali:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,29 @@
...
+
+function average(num1, num2, num3) {
+  const avg = (num1 + num2 + num3) / 3;
+  console.log(avg);
+}
+average(Number(args[2]), Number(args[3]), Number(args[4]));
+
+//minimist
+
+const minimist = require("minimist");
```
margaritluch, line 16: It's a good practice to put all the import statements in the top of the file. It makes the code more readable :)
So I will suggest to move this line up.

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,29 @@
...
+const require = createRequire(import.meta.url);
+
+//process
+
+const args = process.argv;
+
+function average(num1, num2, num3) {
+  const avg = (num1 + num2 + num3) / 3;
+  console.log(avg);
+}
```
margaritluch, line 11: What if the input is not a number but a string or something else? I will suggest to make an if statement that checks that the input is number and if not , then it tells the user to use number instead :)

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,29 @@
...
+average(Number(args[2]), Number(args[3]), Number(args[4]));
+
+//minimist
+
+const minimist = require("minimist");
+
+const minimistArgs = minimist(process.argv.slice(2));
+const values = Object.keys(minimistArgs).filter((key) => key !== "_")
```
margaritluch, line 19: Here again I will suggest to handle potential errors. In that way, you will be able to escape bugs. If the argument is a pure string, fx the argument is "apple" then the code will fail without telling what's wrong. So you can solve that problem by an if statement, where you can check if the argument is not number then throw error with an error message.

margaritluch (commented):
It looks great Hadis. I just have some few suggestions 😊

PR #199, opened 2024-02-21
node-week1/gizem

GizemSavci:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,24 @@
+// Function to calculate average of inputs with process
+function averageInput(){
+    const inputs = process.argv.slice(2)
+    let total = 0;
+    if (inputs.length < 3) {
+        console.log('Please type three numbers.');
+    } else {
+        let allNumbers = true;
+        inputs.forEach(input => {
```
julczka, line 9: Do you think you could use another Array method? For example, one that runs a callback function on each element in the array and returns a boolean?

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,24 @@
+// Function to calculate average of inputs with process
+function averageInput(){
+    const inputs = process.argv.slice(2)
+    let total = 0;
+    if (inputs.length < 3) {
+        console.log('Please type three numbers.');
+    } else {
+        let allNumbers = true;
```
julczka, line 8: it is generally safer/more readable to initiate your boolean variables as `false` rather then as `true`

julczka (changes requested):
Nice work, you can try and think how to refactor this function so it has less nested if/else statements, wich is better for readablity. For example you can the [patterns described in this article ](https://dev.to/sylwiavargas/5-ways-to-refactor-if-else-statements-in-js-functions-208e#5-noelsereturns-and-guard-clauses)

PR #200, opened 2024-02-21
node-week1/anastasiia

humenyuk16:
Warmup,  homework for node week 1

### DIFF
--- a/6_nodejs/week1/avg.js
+++ b/6_nodejs/week1/avg.js
@@ -1,0 +1,23 @@
+const args = process.argv.slice(2);
+
+if(args.length === 0){
+    console.log("No arguments provided");
+}else{
+    let sum = 0;
+    let count = 0;
+    for(let i = 0; i < args.length; i++){
+        const num = parseFloat(args[i]);
+        if(!isNaN(num)){
+            sum += num;
+            count++;
+        }else{
+            console.log("Invalid number: " + args[i]);
+        }
+    }
+    if(count > 0){
+        const avg = sum / count;
+        console.log(`Average: ${avg}`);
+    }else{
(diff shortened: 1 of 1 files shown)


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,23 @@
+const args = process.argv.slice(2);
+
+if(args.length === 0){
```
LucasIversen, line 3: Good

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,23 @@
...
+    console.log("No arguments provided");
+}else{
+    let sum = 0;
+    let count = 0;
+    for(let i = 0; i < args.length; i++){
```
LucasIversen, line 8: good for loop

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,23 @@
...
+        const num = parseFloat(args[i]);
+        if(!isNaN(num)){
```
LucasIversen, line 10: good check 👌

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,23 @@
+const args = process.argv.slice(2);
+
+if(args.length === 0){
+    console.log("No arguments provided");
+}else{
+    let sum = 0;
+    let count = 0;
```
LucasIversen, line 7: It might be a bit overkill to handle a count, instead of just using args.length, but with your isNaN check, i see how it makes sense. You could also use `args.filter((arg) => !isNan(arg)).length`

LucasIversen (approved):
Good job 👌

PR #201, opened 2024-02-22
databases-week3/olesia

AlesyaMazurenko:


MercedesUbeira (approved):
Hey! Really great job! The only suggestion I have for you is to use AUTO_INCREMENT in the ID of all the tables will help you, so every time you add something to the table, the id gets automatically generated :)

So when you do this:

`INSERT INTO review (`id`, `title`, `description`, `meal_id`, `stars`, `created_date`)
VALUES 
(1, 'meal quality', 'The quality of material, taste and freshness was excellent', 1, 10, '2024-02-01'),` the ID will be created automatically and you don't need to add it manually like you did here :)

PR #206, opened 2024-02-23
node-week1/sanazy

Sanazre:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,8 @@
+let avg = 0;
+for (let j = 2; j < process.argv.length; j++) {
+  const a = parseInt(process.argv[2]);
+  const b = parseInt(process.argv[3]);
+  const c = parseInt(process.argv[4]);
+  avg = (a + b + c) / 3;
```
slastushevskyi, line 6: But what's gonna happen if the user enters e.g. a string? 
It would be nice if we created a guard, for instance: 

  if (isNaN(a) || isNaN(b) || isNaN(c)) {
    console.log("Error: Please provide three numbers");
    process.exit(1);
  }

slastushevskyi (commented):
Hey, Sanaz 👋
You are using a nice approach in this task. 
I've left one suggestion for you 🙂

PR #212, opened 2024-02-27
node-week1/olesia

AlesyaMazurenko:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,20 @@
+import { createRequire } from "module";
+const require = createRequire(import.meta.url);
```
kiko-bkr, line 2: These imports are not needed, as you do not need to use any libraries for the exercise. Removing these lines your code runs perfectly. 😄

6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,20 @@
...
+    let sum = 0;
+    for (let i = 2; i < args.length; i++) {
+        sum += +args[i];
+    }
+
+    if (sum > 0) {
+        const avg = sum / (args.length - 2);  
+        console.log(avg);
+    } else {
+        console.log('Please insert numbers');
```
kiko-bkr, line 16: Well done on handling the case when the user provides no arguments, and letting the user know what they did wrong. ⚡

kiko-bkr (changes requested):
Well written! Remove the first 2 lines and you are done. 😃

PR #215, opened 2024-02-28
node-week2/Hadis

Hadis-jamali:


6_nodejs/week2/app.js
```diff
@@ -0,0 +1,72 @@
...
+    res.json(documents);
+  } else {
+    const filteredDocuments = documents.filter((doc) => {
+      return Object.values(doc).some((value) =>
+        value.toString().toLowerCase().includes(query.toLowerCase())
+      );
+    });
+    res.json({ data: filteredDocuments, message: "ok" });
+  }
+});
```
hereisamin: Good job, as it has only one condition, I would write is as ternary syntax and make it short like following:
```
  const searchResult = query ? data.filter(each => Object.values(each).join(' ').search(query) > -1 ) : data
  
  res.json({ data: searchResult, message: "ok" });
``` 
Notes:
- it's best practice to do the response once at the end
- you don't need to send `no result` message in here, as you send an empty array and the frontend will handle the rest.
- using `search` method in js will be good practice, it returns either the index of search key or -1 as not found

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,72 @@
...
+
+//POST /search
+
+app.post("/search", (req, res) => {
+  const query = req.query.q;
+  const fields = req.body.fields;
+
+  if (query && fields) {
+    res.status(400).send("Bad Request: Both q and fields cannot be provided");
+    return;
```
hereisamin, line 42: `return res.status(400).send("Bad Request: Both q and fields cannot be provided");`

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+});
+
+//POST /search
+
+app.post("/search", (req, res) => {
+  const query = req.query.q;
+  const fields = req.body.fields;
+
+  if (query && fields) {
```
hereisamin: If you extend this if then you can get rid of the last else in line `63`
`if ((query && fields) || (!query && !fields) )`

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,72 @@
...
+  } else if (fields) {
+    filteredDocuments = documents.filter((doc) => {
+      return Object.entries(fields).every(
+        ([field, value]) => doc[field] && doc[field].toString() === value.toString()
+      );
+    });
+  } else {
+    res.status(400).send("Bad Request: q or fields must be provided");
+    return;
+  }
```
hereisamin:
```
const filteredDocuments = documents.filter((doc) => {
if (query) {
 return data.filter(eachData => Object.values(eachData).join(' ').search('hello') > -1)
} else {
return Object.entries(fields).every(([field, value]) => doc[field] && doc[field].toString() === value.toString());
}
    });
```

hereisamin (changes requested):
Good job Hadis 👍

6_nodejs/week2/app.js
Hadis-jamali: done

6_nodejs/week2/app.js
Hadis-jamali: great syntax. thanks alot

PR #216, opened 2024-02-28
node-week2/Niloufar

Niloufar97:


6_nodejs/week2/app.js
```diff
@@ -0,0 +1,87 @@
...
+        (value) => typeof value === "string" && value.includes(query)
+      );
+    });
+    if (filteredData.length === 0) {
+      return res.send("not found");
+    }
+    return res.json({ data: filteredData, message: "ok" });
+  }
+  res.json({ data: data, message: "ok" });
+});
```
hereisamin: Good job, as it has only one condition, I would write is as ternary syntax and make it short like following:
```
  const searchResult = query ? data.filter(each => Object.values(each).join(' ').search(query) > -1 ) : data
  
  res.json({ data: searchResult, message: "ok" });
``` 
Notes:
- it's best practice to do the response once at the end
- you don't need to send `no result` message in here, as you send an empty array and the frontend will handle the rest.
- using `search` method in js will be good practice, it returns either the index of search key or -1 as not found

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,87 @@
...
+
+// GET /documents/:id
+app.get("/search/:id", (req, res) => {
+  const dataId = +req.params.id;
+  const myData = data.find((data) => data.id === dataId);
+  if (!myData) {
+    return res.status(404).json({ data: null, message: "not found" });
+  }
+  res.json({ data: myData, message: "ok" });
```
hereisamin: This is right and clean but if you want to follow one response at end rule it could be this way:

```
const result = data.find((data) => data.id === +req.params.id)
  const response = result.length ? { status: 200, message: 'ok' } : { status: 404, message: 'not found' }
  
  res.status(response.status).json({ data: response.message, message: response.message })
```

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,87 @@
...
+      });
+    });
+    
+    if (dataWithField.length === 0) {
+        return res.status(404).json({ data: null, message: "not found" });
+    }
+    return res.status(200).json({ data: dataWithField, message: "ok" });
+} else {
+    return res.json({ data: data, message: "ok" });
+  }
```
hereisamin: This block, which handles `POST` for `/search` is a bit messy, I suggest you do a cleanup and try to use what I mentioned as practice and ask me for another review :)

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,87 @@
...
+  } else if (fields) {
+
+    // it checks if the key is exsit in my data
+
+    // const keys = Object.keys(fields);
+    // const dataWithField = data.filter(item => {
+    //     return keys.every(key => {
+    //         return item.hasOwnProperty(key);
+    //     });
+    // });
```
hereisamin: To make it more professional, try not to include code comments before sending the code for final review.

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,87 @@
...
+  if (!myData) {
+    return res.status(404).json({ data: null, message: "not found" });
+  }
+  res.json({ data: myData, message: "ok" });
+});
+
+// POST /search
+app.post("/search", (req, res) => {
+  const query = req.query.q;
+  console.log(req.body.fields);
```
hereisamin: To make it more professional, try to clean console.log(s) before sending the code for final review.

hereisamin (approved):
Good job 👍

6_nodejs/week2/app.js
Niloufar97: Thank you so much for reviewing my PR. I tried to fix the problems that you mentioned. but for the last part, I couldn't make it very well. I would appreciate it if you could help me with that.

PR #218, opened 2024-02-29
Nodejs-week1/anjali

anjali744:


6_nodejs/week1/avg.js
```diff
@@ -0,0 +1,13 @@
...
+  for (let i = 0; i < numbers.length; i++) {
+    sum += parseInt(numbers[i]);
+  }
+  return sum / numbers.length;
+}
+
+//retrieves the command line arguments
+const args = process.argv.slice(2);
+const average = calculateAverage(args);
+console.log(average);
```
kiko-bkr, line 13: Well implemented! :zap:

6_nodejs/week1/avg.js
```diff
@@ -8,6 +8,30 @@ function calculateAverage(numbers) {
...
+const arguments = process.argv.slice(2);
+const average = calculateAverage(arguments);
 console.log(average);
+
+//--------Optional question--------------------------------------
+//arguments are mix of numbers and string
+function printNumbersAndStrings(args) {
+  if (args.length === 0) {
+    console.log("There is no arguments.");
+    return;
```
kiko-bkr, line 20: Great thing to check!

6_nodejs/week1/avg.js
```diff
@@ -8,6 +8,30 @@ function calculateAverage(numbers) {
...
+  for (let i = 0; i < args.length; i++) {
+    const num = parseFloat(args[i]);
+    if (!isNaN(num)) {
+      numbers.push(num);
+    } else {
+      strings.push(args[i]);
+    }
+  }
+  console.log("Numbers:", numbers);
+  console.log("Strings:", strings);
```
kiko-bkr, line 33: This is one way to handle when the user accidentally provides strings and numbers. However, as the purpose of the code is to calculate the average of **numbers**, we shall treat inputs that contain **strings** as **incorrect inputs**. Thus, if the user inputs a string as an argument, we should handle it as an **error** and let the user know their mistake.

kiko-bkr (commented):
Well done! I appreciate that you went out of your way to also solve the additional optional part! I suggest you rewrite the code to handle string inputs as errors, but other than that great job! 🏆

PR #222, opened 2024-03-01
node-week2/katayoun

katayouny:


6_nodejs/week2/app.js
```diff
@@ -0,0 +1,110 @@
...
+    // filter documents' objects, as doc, based on whether any field matches the query string
+    const filteredDocuments = documents.filter((doc) =>
+      Object.values(doc).some(
+        (value) => typeof value === "string" && value.includes(query)
+      )
+    );
+    if (filteredDocuments) {
+      res.json(filteredDocuments);
+    } else {
+      res.send("no match found");
```
julczka, line 43: you might want to keep the endpoint response consistent - so if for a certain query, the response is a JSON with an array of objects, then if there is nothing found for a different query the endpoint should return a JSON with an empty array, not a string instead of JSON. That way you keep the type of data that the endpoint returns the same - it will always be a JSON and that will be easier to handle for anyone who is using your API.

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+});
+//----------------------------------------------------------------
+//GET /documents/:id
+//find and respond with the document matching the id parameter.
+//If there is no such document, respond with a 404 Not Found.
+//You can assume that the document IDs are unique so there's no need to handle duplicates.
+app.get("/documents/:id", (req, res) => {
+  console.log(req.params);
+  const documentId = +req.params.id;
```
julczka, line 54: I suppose you have a plus here because `id` in the JSON file is a number. In general ID's are strings, so the plus here would cause you some trouble. In the "real world" an ID is very often something called GUID. You can check out this generator https://www.guidgenerator.com/ . It contains both numbers and letters and it is so long that it guarantees that the combination will always be unique. It is always safer to assume that ID is a string.

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,110 @@
...
+//fields is an object where it will be possible to filter by specific fields.
+app.post("/search", (req, res) => {
+  const query = req.query.q; // Extract the 'q' query parameter from the request URL .Example: http://localhost:3000/search?q=50
+  const fields = req.body.fields; // Extract the 'fields' object from the request body . Example: http://localhost:3000/search?fields=price
+  console.log(query);
+  console.log(fields);
+  // Check if both query parameter and fields in the body are provided
+  if (query && fields) {
+    res.status(400).json({
+      error: "Both 'q' parameter and 'fields' in body cannot be provided",
```
julczka, line 79: Nice anc clean error messge! a big plus!

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,110 @@
...
+//----------------------------------------------------------------
+//POST /search
+//This endpoint is sort of like GET /search.
+//It also accepts a query parameter called q and it should behave just like GET /search.
+//But it also accepts a field called fields in the JSON request body.
+//fields is an object where it will be possible to filter by specific fields.
+app.post("/search", (req, res) => {
+  const query = req.query.q; // Extract the 'q' query parameter from the request URL .Example: http://localhost:3000/search?q=50
+  const fields = req.body.fields; // Extract the 'fields' object from the request body . Example: http://localhost:3000/search?fields=price
+  console.log(query);
```
julczka, line 74: You might want to remove console.logs that were made for debugging purposes before submitting a PR :)

julczka (approved):
Very nice clean code, good job!

katayouny:
Hey @julczka 
Thank you so much for the review and suggestions :)

PR #223, opened 2024-03-01
node-week2/Helena

helena69hme:


6_nodejs/week2/app.js
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+        const filteredDocuments = documents.filter((doc) => {
+            return Object.values(doc).some((value) =>
+              value.toString().includes(query)
+            );
+          });
+          if (filteredDocuments) {
+          res.json(filteredDocuments);
+        } else {
+          res.send("Unable to found");
```
julczka, line 28: you might want to keep the endpoint response consistent - so if for a certain query the response is a Json with an array of objects, then if there is nothing found for a different query the endpoint should return a json with empty array, not a string instead of json. That way you keep the type of data that endpoint return the same - it will always be a json and that will be easier to handle for anyone who is using your API.

julczka (approved):
Code looks nice and clean! Good job!

PR #224, opened 2024-03-02
node-week2/sanaz

Sanazre:


6_nodejs/week2/app.js
```diff
@@ -0,0 +1,87 @@
...
+    // filter documents' objects, as doc, based on whether any field matches the query string
+    const filteredDocuments = documents.filter((doc) =>
+      Object.values(doc).some(
+        (value) => typeof value === "string" && value.includes(query)
+      )
+    );
+    if (filteredDocuments) {
+      res.json(filteredDocuments);
+    } else {
+      res.send("no match found");
```
julczka, line 29: I think in this case it would be better if the endpoint returns an empty array instead of a string. Imagine receiving this data in the client: one time you get an array of objects, the other time you get a string and you have two handle both of those cases. It would be way earlier just to get an empty array if there are no search results found.

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,87 @@
...
+});
+//part2:
+app.get("/documents/:id", (req, res) => {
+  console.log(req.params);
+  const documentId = +req.params.id;
+  //console.log(documentId);
+  const resultDoc = documents.find((document) => document.id === documentId);
+  if (resultDoc) {
+    res.json(resultDoc);
+    return;
```
julczka, line 41: No need fro both `return` and `else` statement. In this case if you use return inside an if block you do not need else block - similarly you do not need to return from if block if there is nothing more in this function.

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,87 @@
...
+  } else {
+    res.status(404).json({ error: "Document not found" });
+  }
+});
+
+//part3:
+app.post("/search", (req, res) => {
+  const query = req.query.q;
+  const fields = req.body.fields;
+  console.log(query);
```
julczka, line 51: It is a good practice to remove unnecessary console log statements from PR before submitting it :)

julczka (approved):
Nice work Sanaz!

PR #226, opened 2024-03-05
node-week3/hadis

Hadis-jamali:


6_nodejs/week3/app.js
```diff
@@ -0,0 +1,51 @@
...
+
+const contactsAPIRouter = express.Router();
+apiRouter.use("/contacts", contactsAPIRouter);
+
+contactsAPIRouter.get("/", async (req, res) => {
+  let query = knexInstance.select("*").from("contacts");
+
+  if ("sort" in req.query) {
+    const orderBy = req.query.sort.toString();
+    const validColumns = ["first_name", "last_name", "email"];
```
kiko-bkr, line 31: A great way to prevent SQL injection is **allowlisting**, as you demonstrate here. ⚡

6_nodejs/week3/app.js
```diff
@@ -0,0 +1,51 @@
...
+
+    if (orderBy.length > 0 && validColumns.includes(orderBy)) {
+      query = query.orderBy(orderBy, req.query.order === "desc" ? "desc" : "asc");
```
kiko-bkr, line 34: You separate the ordering choice to a different query parameter. Good choice! 📱 Maybe a good idea would have been to let the user know if they provided an invalid ordering choice.

kiko-bkr (approved):
Well prevented! 🏅

PR #227, opened 2024-03-05
node-week3/anastasiia

humenyuk16:


6_nodejs/week3/app.js
```diff
@@ -0,0 +1,49 @@
...
+app.use("/api", apiRouter);
+
+const contactsAPIRouter = express.Router();
+apiRouter.use("/contacts", contactsAPIRouter);
+
+//Preventing SQL Injection
+contactsAPIRouter.get("/", async (req, res) => {
+    try{
+        const { sort } = req.query;
+        const allowedSortKeys = ["id", "first_name", "last_name", "email", "phone"];
```
kiko-bkr, line 30: You solved the issue of SQL injection vulnerability via [allowlisting](https://planetscale.com/blog/how-to-prevent-sql-injection-attacks-in-node-js#allowlisting). Great! 👍 One thing: the original code allowed the user to determine the sorting direction: either `ASC` or `DESC`. You took away this feature, but you can easily re-introduce it via an extra query parameter, which you should also allowlist.

kiko-bkr (changes requested):
Well done! One small note about the functionality of the solution, but great code overall. ⚡

PR #228, opened 2024-03-05
node-week2/gizem

GizemSavci:


6_nodejs/week2/app.js
```diff
@@ -0,0 +1,62 @@
+import express from 'express'
```
sergii-borodin, line 1: Seems you forgot to import the local JSON file with data.

Import the JSON file you can do like this:
import fs from "fs";
const data = JSON.parse(fs.readFileSync("./document.json"));

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,62 @@
...
+        res.json(filteredData); 
+    } else {
+      res.json(data); 
+    }
+})
+
+// GET /documents/:id
+app.get('/documents/:id', (req, res) => {
+  const id = parseInt(req.params.id);
+  const data = documents.find(item => item.id === id);
```
sergii-borodin, line 32: In this route, you are working on an array called documents but in the first one, you call it data.

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,62 @@
...
+
+// POST /search
+app.post('/search', (req, res) => {
+  const query = req.query.q;
+  const fields = req.body.fields;
+
+  if (query && fields) {
+    res.status(400).send("q and fields cannot be provide at the same time.")
+  } else if (query) {
+    const filteredData = data.filter(item =>{
```
sergii-borodin, line 48: In this route, I found a few things to improve.
Create a flag variable which by default is equal to false and I would call it something like isFound before filtering data.
While filtering the array in case adding an item matches a search, change it to be true.
After the filtering, you can check if the isFound variable is still equal to false then you can give feedback that will indicate a successful or unsuccessful search. I will provide an example below.

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,62 @@
...
+      for (const key in item) {
+        if (item[key].toString().includes(query)){
+          return true;
+        }
+      } return false;
+    });
+    res.json(filteredData);
+}
```
sergii-borodin, line 56: According to the task, you forgot to implement another case where a field called fields in the JSON request body was not provided.

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,62 @@
...
+});
+
```
sergii-borodin, line 58: Here is an example of how I would implement this route:
```
app.post("/search", (req, res) => {
  const query = req.query.q;
  const fields = req.body.fields;

  if (!query && !fields) {
    res.status(400).send("Please provide searching parameters");
    return;
  }

  if (query && fields) {
    res.status(400).send("q and fields cannot be provide at the same time.");

    return;
  } else if (query) {
    let isFound = false;
    const filteredData = data.filter((item) => {
      for (const key in item) {
        if (item[key].toString().includes(query)) {
          isFound === false && (isFound = true);
          return true;
        }
      }
      return false;
    });
    if (!isFound) {
      res.status(200).send("No match found");
    } else {
      res.json(filteredData);
    }
  } else if (fields) {
    let isFound = false;
    const [searchingKey, searchingValue] = Object.entries(fields)[0];
    const filteredData = data.filter((item) => {
      for (const key in item) {
        if (key === searchingKey && item[key] === searchingValue) {
          isFound === false && (isFound = true);
          return true;
        }
      }
      return false;
    });
    if (!isFound) {
      res.status(200).send("No match found");
    } else {
      res.json(filteredData);
    }
  }
});

```

sergii-borodin (changes requested):
Hello Gizem!
Sorry for the delay in checking your homework!
I found some issues in your code.
It gives me a reason to think you didn't test your code. I want to give you advice to check your code before submitting homework. Then you can see how well you understood the material you went through.

Feel free to let me know if you have any questions or need help. 
Your second part of the homework I will check ASAP

PR #229, opened 2024-03-05
node-week2/yousseftopaji

yousseftopaji:
This is the Warmup part of the homework.

I have added module: type to package.json, but then it gave me an error 
ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and 'C:\Users\youss\HYF\nodejs-week1\class27-homework\6_nodejs\week2\package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.    

And therefore I deleted it, and it works fine!
Could you please just mention why that happened?

### DIFF
--- a/6_nodejs/week2/app.js
+++ b/6_nodejs/week2/app.js
@@ -1,0 +1,84 @@
+const express = require("express");
+const app = express();
+const fs = require("fs");
+const port = 3000;
+
+app.use(express.json());
+
+const documents = JSON.parse(fs.readFileSync("documents.json", "utf-8"));
+
+app.get("/", (req, res) => {
+  res.send("Hi there");
+});
+// GET /search
+app.get("/search", (req, res) => {
+  const query = req.query.q;
+
+  if (!query) {
+    res.json(documents);
+  } else {
+    const results = documents.filter((doc) => {
+      for (const key in doc) {
+        if (doc[key].toString().includes(query)) {
+          return true;
+        }
+      }
+      return false;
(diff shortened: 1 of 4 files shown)


6_nodejs/week2/app.js
```diff
@@ -0,0 +1,84 @@
...
+        if (doc[key].toString().includes(query)) {
+          return true;
+        }
+      }
+      return false;
+    });
+    res.json(results);
+  }
+});
+
```
MaherHussain, line 31: good job, since we have an array method that helps us to pass conditions let's use it 
```
`app.get("/search", (req, res) => {

  const results = documents.filter((docs) =>
    Object.values(docs).some((value) => value.toString().toLowerCase().includes(query.toLocaleLowerCase())))

  query ? res.json(results) : res.json(documents);
});`
```

6_nodejs/week2/app.js
```diff
@@ -0,0 +1,84 @@
...
+            return true;
+          }
+        }
+        return false;
+      });
+      res.json(results);
+    } else if (fields) {
+      const results = documents.filter((doc) => {
+        for (const key in fields) {
+          if (doc[key] && doc[key].toString() === fields[key]) {
```
MaherHussain, line 67: no need to check  doc[key] here

MaherHussain:
All in all, you did a good job, Youssef

PR #231, opened 2024-03-07
node-week3/katayoun

katayouny:


6_nodejs/week3/app.js
```diff
@@ -0,0 +1,58 @@
...
+  let query = knexInstance.select("*").from("contacts");
+
+  if ("sort" in req.query) {   
+    const orderBy = req.query.sort.toString(); // converting the sort to string to make it able to get used by .orderByRaw('')
+    const allowedColumnsForSort = [
+      "first_name",
+      "last_name",
+      "email",
+      "phone", 
+    ];
```
kiko-bkr, line 39: You have solved the SQL injection vulnerability via [allowlisting the paremeter](https://planetscale.com/blog/how-to-prevent-sql-injection-attacks-in-node-js#allowlisting). Great job! 👍 This is necessary since we as software engineers can **NEVER** trust user input. 🤞

6_nodejs/week3/app.js
```diff
@@ -0,0 +1,58 @@
...
+    if (orderBy.length > 0 && allowedColumnsForSort.includes(orderBy)) {// Check if the orderBy parameter is inside the allowedColumnsForSort
+      query = query.orderByRaw(orderBy); 
+    }
```
kiko-bkr, line 42: Your solution works fine, but it has removed some of the original functionality. The original code allowed the user to determine the sort direction, but they are no longer able to do that. This is easy to fix though, as you can just add a new query parameter called fx. `orderDirection` or `orderDir`. REMEMBER though: each new parameter is a new SQL injection possibility; make sure to allowlist the new paramater too! 🧠

kiko-bkr (commented):
Nice solution and proof of the vulnerability! 🏅  Your solution removed some of the original functionality, but that can be easily fixed. Refer to the comments for details. 😊

PR #237, opened 2024-03-10
node-week3/Viki

VictoriaDem17:


6_nodejs/week3/app.js
```diff
@@ -0,0 +1,65 @@
...
+
+// SQL injection cheat sheet - https://portswigger.net/web-security/sql-injection/cheat-sheet
+
+contactsAPIRouter.get("/", async (req, res) => {
+  let query = knexInstance.select("*").from("contacts");
+
+  if ("sort" in req.query) {
+    const orderBy = req.query.sort.toString();
+    // preventing SQL injection using parameterized queries instead of string concatenation
+    const preparedStatementQueries = [
```
NVedsted, line 41: Clever solution! 🧠 However, we're missing some functionality: you are not able to specify whether it is ascending or descending! 😮 Can you think of a way to support this without introducing a security issue?

NVedsted (changes requested):
Nice solution! ⭐ There is a little regression in the functionality that I have left a comment about. 😄

PR #240, opened 2024-03-10
node-week2/anjali

anjali744:


6_nodejs/week2/app.js
```diff
@@ -0,0 +1,68 @@
...
+
+// GET /search?id=2
+app.get("/search", (req, res) => {
+  const query = req.query.q;
+  if (!query) {
+    res.json(documents);
+  } else {
+    const results = documents.filter((doc) =>
+      Object.values(doc).some(
+        (value) => typeof value === "string" && value.includes(query)
```
MahmoudYounes: This is not a requirement that is mentioned in the homework so it is a nitty pick. value.includes is case sensitive so if I call this endpoint with apple instead of Apple I will get nothing. how about we make this case insensitive. i.e, value.toLowerCase().includes(query)? :)

MahmoudYounes:
Very Nice job!

MahmoudYounes (approved):

anjali744:
I resolve it and making search and query case-insensitive.😊

PR #243, opened 2024-03-11
react1-week1/hadis

Hadis-jamali:


7_react/react1/week1/ToDoList/src/App.jsx
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from "react";
```
BrantalikP: To keep the codebase clean as much as possible you should always remove unused imports. You can remove the `React` import as well as it is not required anymore in the new React versions :)

BrantalikP (changes requested):
Well done @Hadis-jamali  👏, just please remove the unused imports :)

7_react/react1/week1/ToDoList/src/App.jsx
Hadis-jamali: done

BrantalikP (approved):
🔥

PR #245, opened 2024-03-11
react1-week1/Viki

VictoriaDem17:
Todo_list_app by React:
week1: static todo list with three items; components.


7_react/react1/week1/todo_list_app/src/App.js
```diff
@@ -0,0 +1,36 @@
+import logo from "./logo.svg";
+import "./App.css";
+import { TodoList } from "./todoList/TodoList";
+
+//React1-week1
+// -render a basic static todo list with three items
+// -for each item render a description and a deadline date
+
+function App() {
+  const todos = [
```
BrantalikP, line 10: maybe it's too early to discuss the lifestyle ecosystem in React, but long story short leaving the variable with static data within render method (inside of the React Component) is not a good practice. That being said, the fix is simple. just move it above the `function App() {` so it will be 
```
 const todos = [...your data]
 function App() {
```
We can talk about it more, if you are interested :)

BrantalikP (approved):
Besides that one thing everything looks great. Good job @VictoriaDem17 👏

PR #247, opened 2024-03-12
node-week3/Helena

helena69hme:


6_nodejs/week3/app.js
```diff
@@ -0,0 +1,57 @@
...
+
+  if ("sort" in req.query) {
+    const orderBy = req.query.sort;
+    const allowedColumnsForSort = [
+        "first_name",
+        "last_name",
+        "email",
+        "phone",
+        "id"
+      ];
```
kiko-bkr, line 38: You decided to fix the SQL injection vulnerability via [allowlisting the parameters](https://planetscale.com/blog/how-to-prevent-sql-injection-attacks-in-node-js#allowlisting). It is a great way to handle user input, since as we know, user input should **NEVER** be trusted. 🧠

6_nodejs/week3/app.js
```diff
@@ -0,0 +1,57 @@
...
+      if (orderBy.length > 0 && allowedColumnsForSort.includes(orderBy)) {
+        query = query.orderByRaw(orderBy); 
+      }
```
kiko-bkr, line 41: Your solution works just fine for letting the user decide which column to sort the result on. However, in the original code, the user also had the option to determine whether the sorting should be ascending or descending using the second part of the original parameter.

Currenlty your new solution removed this functionality. We prefer to never sacrifice functionality. Thankfully, you can easily add this functionality back to your code by having another query parameter, let's say `orderDir`. Remember though, new parameter = new SQL injection possibility! Make sure to allowlist that parameter too. 😉 🛑

kiko-bkr (changes requested):
Well done! The solution is nice and simple, yet it removes some of the original functionality. Refer to my comments for details.

Moreover: could you provide a screenshot, or at the very least a text file of a request link that would perform SQL injection on the **original** code? I cannot see such a file, and the assignment requested it. 😊 Essentially just come up with query parameters that would drop the `contacts` table.

PR #248, opened 2024-03-12
react1-week1/anastasiia

humenyuk16:
todo list


7_react/react1/week1/vite-project/src/App.jsx
```diff
@@ -0,0 +1,35 @@
...
+      deadline: "05-04-2024",
+    },
+    {
+      description: "Finish homework for React1",
+      deadline: "15-03-2024",
+    },
+  ]);
+
+  const addTodo = (newTodo) => {
+    setTodos([...todos, {...newTodo, id:  todos.length + 1}]);
```
BrantalikP, line 23: Well done, just one small thing, although your approach is not wrong, due to `useState` is an asynchronous action, I prefer using prevData directly from the set method like this:
setTodos(prevTodos => [...prevTodos, { id: todos.length + 1, ...newTodo }]);. Otherwise, if there were any disruptions in between, the data could be allocated incorrectly.

7_react/react1/week1/vite-project/src/components/TodoList.jsx
```diff
@@ -0,0 +1,20 @@
+
+import React, { useState } from 'react';
```
BrantalikP, line 2: as the codebase should be always clean as much as possible, there should be hanging any unused imports, please remove any redundant code you will find unnecessary :) Also React import is not required anymore in the newer versions of React, you can safely remove it

BrantalikP (changes requested):
Amazing job @humenyuk16 👏 I like the extra work you did for the assignment, just minor suggestion, other than that. its great 👍

PR #251, opened 2024-03-14
react1-week1/Ihor

IhorKulytskyi:
Homework


7_react/react1/week1/my-app/src/App.jsx
```diff
@@ -0,0 +1,32 @@
+import './App.css';
+import TodoList from './Components/TodoList/TodoList';
+
+function App () {
+  const todos = [
```
SvitlanaBurlaka, line 5: i would recommend to move your array"todos" outside of function App. :) You cac e.g create folder 'data' with data.jsx and have there your  `const todos = []` and then import it where you need.

SvitlanaBurlaka (commented):
Looks good! Just was trying to run the project locally, but could not find package.file. Maybe you did not push it ?

PR #252, opened 2024-03-15
node-week3/yousseftopaji

yousseftopaji:
In the attached video I implement a SQL injection.

And in the app.js I implement a method to prevent SQL injection.

### DIFF
--- a/6_nodejs/week3/app.js
+++ b/6_nodejs/week3/app.js
@@ -1,0 +1,53 @@
+import knex from "knex";
+const knexInstance = knex({
+  client: "mysql2",
+  connection: {
+    host: process.env.DB_HOST || "127.0.0.1",
+    port: process.env.DB_PORT || 3306,
+    user: process.env.DB_USER || "root",
+    password: process.env.DB_PASSWORD || "pass",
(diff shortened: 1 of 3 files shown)


6_nodejs/week3/app.js
```diff
@@ -0,0 +1,53 @@
...
+
+  if ("sort" in req.query) {
+    const orderBy = req.query.sort.toString();
+    if (orderBy === "first_name") {
+      query = query.orderBy(orderBy);
+    } else if (orderBy === "last_name") {
+      query = query.orderBy(orderBy, "desc");
+    } else {
+      res.send("invalid sort");
+    }
```
kiko-bkr, line 37: Well done on the video of the injection attack. 

I see you have fixed the issue by making sure the parameter value is allowed. Good job. 💪  One note is that you did this at the expense of functionality: currently there is no way for the user to determine the ordering direction. First names are always ordered ascending, last names are always ordered descending. You could fix this by implementing another parameter, for example `sortDir`.

kiko-bkr (changes requested):
Great job! One thing to pay attention to is to almost never sacrifice functionality in new implementations. Make sure that the new code works the same as before. Refer to the comment about the suggestion. 🎖️

PR #253, opened 2024-03-15
react1-week1/Helena

helena69hme:


MaherHussain:
Hi Helena.
the homework requirement is creating separate components not all in App components. and pass props to it from the parent component.

BrantalikP (changes requested):
Hey @helena69hme, good job with submitting the homework. However, a part of the assignment is to have a functional app, as outlined here: https://github.com/HackYourFuture-CPH/React/blob/main/react1/week1/homework.md. To do so, you need to initialize the React project via `npm create vite@latest`. After that, you can start implementing the components, and you will also be able to run the app via the `npm run dev` command. Keep in mind that components should be separated, as that's the core purpose of React. Good luck!

PR #260, opened 2024-03-19
react1-week2/Niloufar

Niloufar97:


7_react/react1/week2/TodoList/src/App.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export type Todo = {
+  id: number,
+  description : string,
+  deadline: string
+}
+export type  DeleteTodo = (id : number) => void
+
+function App() {
+  const [todos, setTodos] = useState<Todo[]>([    {
```
BrantalikP, line 14: There are two reasons why to move the default data to a separate variable, 
1. Data re-rendering, if you leave the data here, it will be rendered along with each re-render, this feels unnecessary as the original data never changes, right now it might not be an issue but if the default would be much longer (let's say thousands of items) it could have a huge impact on performance. The fix is pretty straightforward tho. Just move the items into a separate variable and out of the React component like this:
```
 const initialTodos = [..you data]
 function App() {
  const [todos, setTodos] = useState<Todo[]>(initialTodos)
```
1. Readability

7_react/react1/week2/TodoList/src/App.tsx
```diff
@@ -0,0 +1,52 @@
...
+    deadline: "Thu Sep 14 2017",
+  },
+  {
+    id: 3,
+    description: "Eat breakfast",
+    deadline: "Fri Sep 15 2017",
+  },])
+  
+  const addRandomTodo = () => {
+    setTodos([...todos , {
```
BrantalikP, line 31: this is fine, but it could cause some issues if there would be some async logic between, it's much safer to use prevData from `set` function directly as 
`setTodos(prevTodos => [...prevTodos, newTodo])`

7_react/react1/week2/TodoList/src/App.tsx
```diff
@@ -0,0 +1,52 @@
...
+      id: todos.length + 1,
```
BrantalikP, line 32: using length as an `ID` is not safe. For example, if I have 3 items with ID 1,2,3, then I will delete the first item ( items 2,3 stay) and now I will add a new item, but something is wrong, because of `todos.length + 1` I will have 2 items with ID 3..., its already bad for React to keep track of the items in the list, but if you decide to delete the last item, it will delete 2 items instead, because of the filter logic is looking for all items with ID 3. 

If you do not have any unique text you could use for example timestamp from Date API :) 
I hope it makes sense, if you have questions we can talk about it more :)

BrantalikP (changes requested):
Good job with the homework, I have found some minor issues tho, please check them out and if you have any questions, please reach out to me :)

PR #261, opened 2024-03-19
react1-week2/angel

AngelHenriettaAboah:
Hey there, it's me and I'm here with another PR! Happy Coding....


7_react/react1/week2/src/components/TodoHeader.jsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+function TodoHeader() {
+  return (
+    <div>
```
BrantalikP: There is no need for the extra `div` element, you can return just h1

7_react/react1/week2/src/components/TodoFooter.jsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+function TodoFooter() {
+  return (
+    <div>
```
BrantalikP: There is no need for the extra `div` element, you can return just `p`

7_react/react1/week2/src/components/NewTodoForm.jsx
```diff
@@ -0,0 +1,34 @@
...
+  const [description, setDescription] = useState("");
+  const [deadline, setDeadline] = useState("");
+
+  const handleSubmit = (e) => {
+    e.preventDefault();
+    if (!description.trim() || !deadline.trim()) return;
+    // Add logic to add new todo
+    setDescription("");
+    setDeadline("");
+  };
```
BrantalikP: none of this logic is used, you can safely delete it all from line 4 to 13 :)

7_react/react1/week2/src/components/TodoList.jsx
```diff
@@ -0,0 +1,135 @@
...
+    this.state = {
+      todos: todos,
+      timer: 0,
+      description: "",
+      deadline: "",
+    };
+  }
+
+  componentDidMount() {
+    this.timerID = setInterval(() => this.tick(), 1000);
```
BrantalikP: Well done on implementing the timer 👏! You are the first so far. It's interesting that you are using `class components` Is there a specific reason for this choice? No worries, it's still correct. However, it's worth noting that `class components` are not as commonly used for this purpose nowadays. Here's a more modern approach using `functional components` and `useEffect`. I highly recommend giving it a look when you have time :) https://legacy.reactjs.org/docs/hooks-effect.html.

7_react/react1/week2/src/components/TodoList.jsx
```diff
@@ -0,0 +1,135 @@
...
+  };
+
+  addRandomTodo = () => {
+    const { description, deadline } = this.state;
+    const newTodo = {
+      id: Date.now(),
+      description: description,
+      deadline: deadline,
+      done: false,
+    };
```
BrantalikP, line 52: Right now you can add also empty fields, you should check first whether the input are valid before inserting them :)

BrantalikP (changes requested):
Well done @AngelHenriettaAboah! I've found some small issues but other than that, very good job! 👏

7_react/react1/week2/src/components/TodoHeader.jsx
AngelHenriettaAboah: corrected already....thanks BrantalikP

7_react/react1/week2/src/components/TodoFooter.jsx
AngelHenriettaAboah: corrected here also.

7_react/react1/week2/src/components/NewTodoForm.jsx
AngelHenriettaAboah: altered already

7_react/react1/week2/src/components/TodoList.jsx
AngelHenriettaAboah: okay, I will definitely take it into consideration.

BrantalikP (approved):
Wow, did not expect. You will be that fast. Well done 👏

PR #263, opened 2024-03-21
react1-week3/angel

AngelHenriettaAboah:
Hi there! Here's another PR from me. Happy Coding!


7_react/react1/week3/src/components/TodoList.jsx
```diff
@@ -0,0 +1,182 @@
...
+    setEditTodoId(id);
+    setEditedDescription(description);
+    setEditedDeadline(deadline);
+  };
+
+  const updateTodo = () => {
+    if (
+      !editedDescription ||
+      !editedDeadline ||
+      new Date(editedDeadline) < new Date()
```
SvitlanaBurlaka, line 82: I would suggest to throw an error  with text ' Pay attention to time' or similar 😄  if I am trying to edit todo and setting the time in the past, for example yesterday. It is a bit confusing in this way.

SvitlanaBurlaka (commented):
Good job!
As per description in homework task you can add some errors in those cases(PSST. same what I mentioned in another comment about date in the past):
Worth considering:

What if the description input is empty?
What if the deadline input is empty?
What if the deadline is before the current day?
What if you edit an item with an empty description?

7_react/react1/week3/src/components/TodoList.jsx
AngelHenriettaAboah: okay, will definitely take this into consideration.

PR #268, opened 2024-03-23
react1-week2/maryam

MaryamAarhus:


7_react/react1/week2/todoList2/src/components/todoList/TodoItem.jsx
```diff
@@ -0,0 +1,42 @@
...
+            {isEditing ? (
+                <button onClick={handleSave}>Save</button>
+            ) : (
+                <button onClick={() => setIsEditing(true)}>Edit</button>
+            )}
+        </li>
+    );
+}
+
+TodoItem.propTypes = {
```
BrantalikP, line 35: first time I am seeing someone use the proptypes in their homework, good job 👏

7_react/react1/week2/todoList2/src/components/Timer.jsx
```diff
@@ -0,0 +1,19 @@
+import React, { useState, useEffect } from 'react';
+
+function Timer() {
```
BrantalikP, line 3: The timer is not used anywhere, you can delete it :)

7_react/react1/week2/todoList2/src/components/todoList/TodoItem.jsx
```diff
@@ -0,0 +1,42 @@
+import React, {useState} from 'react';
+import PropTypes from 'prop-types';
+
+function TodoItem({ description, deadline, onDelete, onSave }) {
+    const [isEditing, setIsEditing] = useState(false);
+    const [editedDescription, setEditedDescription] = useState(description);
+
+    const handleSave = () => {
+        onSave(editedDescription);
```
BrantalikP, line 9: The `save` function is missing a type declaration. Also it's not working right now. Maybe you have forgotten to implement the functionality?

7_react/react1/week2/todoList2/src/components/todoList/TodoList.jsx
```diff
@@ -0,0 +1,73 @@
...
+        if (!newTodoDescription) {
+            setErrorMessage('Description cannot be empty.');
+            return;
+        } else if (!newTodoDeadline) {
+            setErrorMessage('Deadline cannot be empty.');
+            return;
+        } else if (new Date(newTodoDeadline) < new Date()) {
+            setErrorMessage('Deadline cannot be in the past.');
+            return;
+        }
```
BrantalikP, line 21: well done.

7_react/react1/week2/todoList2/src/components/todoList/TodoList.jsx
```diff
@@ -0,0 +1,73 @@
...
+            id: Date.now() 
+        };
+        
+        setTodos([...todos, newTodo]);
+        setNewTodoDescription('');
+        setNewTodoDeadline('');
+    };
+
+    const deleteTodo = (id) => {
+        setTodos(todos.filter(todo => todo.id !== id));
```
BrantalikP, line 37: Although it's not incorrect, when you modify data with the `set` method from `useState`, it's better to use its previous state like this: `setTodos(prevTodos => prevTodos.filter(todo => todo.id !== id));`. The reason is that the `set` method is asynchronous, meaning it's not guaranteed that `todo` will hold the correct data at the time you are modifying it.

7_react/react1/week2/todoList2/src/components/todoList/TodoList.jsx
```diff
@@ -0,0 +1,73 @@
+import React, { useState } from 'react';
+import TodoItem from './TodoItem'; 
+import BorderComponent from './BorderComponents';
+
+function TodoList() {
+    const [todos, setTodos] = useState([]);
+    const [newTodoDescription, setNewTodoDescription] = useState('');
+    const [newTodoDeadline, setNewTodoDeadline] = useState('');
```
BrantalikP, line 8: well done with the date time picker! This is just suggestion, but feel free to use it if you want. 

1. You can define the current date by something like this
```
const today = new Date().toISOString().split("T")[0];
function TodoList() {
 const [newTodoDeadline, setNewTodoDeadline] = useState(today);
```
Thanks to that the current day will be always as default value
2. set the current value as `min` date for the input
```
 <input
                type="date"
                min={today}
                value={newTodoDeadline}
```
Thanks to this, you do not need to check whether the selected date is in the past, because the picker won't let you pick the date :) I think this could improve the User experience a bit :)

BrantalikP (changes requested):
Well done! @MaryamAarhus 👏 I really like the app, although I left more comments, just fixing the "onSave" method is required, other things are just suggestions:) 

Just a note: next time please use the correct folder, right now you are using "week2" folder for the `week3` assignment. Thank you!

7_react/react1/week2/todoList2/src/components/todoList/TodoItem.jsx
MaryamAarhus: Thanks a lot for your heartwarming comments 😊 I highly appreciate it!

PR #269, opened 2024-03-23
react1-week2/anjali

anjali744:


7_react/react1/week2/TODO-LIST/src/App.jsx
```diff
@@ -0,0 +1,81 @@
+import React, { useState } from "react";
+import toDoItems from "./component/TodoItems";
+import Timer from "./component/timer";
+import "./App.css";
+
+
+function App() {
+  const [toDoItem, setToDoItems] = useState(toDoItems);
```
kiko-bkr, line 8: The name of the state variable should be `toDoItems`, if the name of the setter is `setToDoItems`. The reason why you did not do this is that the initial to-do array that you import is already called `toDoItems`. However, you can just call the imported initial to-dos as `initialToDoItems`, and then this line can be:

```javascript
const [toDoItems, setToDoItems] = useState(initialToDoItems);
```

7_react/react1/week2/TODO-LIST/src/App.jsx
```diff
@@ -0,0 +1,81 @@
...
+// Function to add a random item
+  const addRandomItem = () => {
+    const randomId = Math.floor(Math.random() * 1000);
+    const newItem = {
+      id: randomId,
+      description: `Task ${randomId}`,
+      deadline: "Some deadline", 
+      done: false // Initially task is not done
+    };
+    setToDoItems([...toDoItems, newItem]);
```
kiko-bkr, line 18: Since `toDoItems` is the const array that you imported, this method would always just set the new state to be the original array plus the **latest** new to-do. However, as noted in the previous comment, if you just make sure that `toDoItems` is the actual state variable, this code will be perfectly fine! 🧠

7_react/react1/week2/TODO-LIST/src/App.jsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <div className="container">
+      <h2>Todo List</h2>
+      <Timer />
+      <button onClick={addRandomItem}>Add Items Here</button>
+      {toDoItems.length === 0 ? (
+        <p>No items</p>
+      ) : (
+      <ul>
+        {toDoItems.map((todos,index) => (
```
kiko-bkr, line 51: I will note one last time why it is important for you to make the first comment's correction. In your code currently `toDoItems` is the constant array that you import. And in this call, you are rendering the elements of that array. However, this array never changes, since it is not the state variable! This is why no matter what you click in the current version of the code, the state stays the same. That must have been frustrating. 😄 

Again, just by making the first comment's correction, this will also work fine.

P.S: You do not actually use the array `index` for anything, so you can just remove that parameter from the callback of `map()`. Moreover, since you are iterating over every individual todo, the other parameter should rather be called `todo`, instead of `todos`. 😉

7_react/react1/week2/TODO-LIST/src/App.jsx
```diff
@@ -0,0 +1,81 @@
...
+          <li key={todos.id}>
+          {/* { <ToDoListItem key={index}  todosId={todos.id} todosDescription={todos.description} todosDeadline={todos.deadline}   /> } */}
+          
+              <span style={{ textDecoration: todos.done ? "line-through" : "none" }}>{todos.description}</span>
+              <input
+                type="checkbox"
+                checked={todos.done}
+                onChange={!todos.done ? markAsDone : markAsNotDone}
```
kiko-bkr, line 59: This is a nice way to pass the functions as callbacks. However, there is one issue: these functions expect the id of the todo! This means you should instead write:
```javascript
onChange={!todo.done ? () => markAsDone(todo.id) : () => markAsNotDone(todo.id)}
```

7_react/react1/week2/TODO-LIST/src/component/timer.jsx
```diff
@@ -0,0 +1,19 @@
...
+            setSeconds(seconds => seconds + 1);
+        }, 1000);
+        return () => clearInterval(interval);
+    }, []);
+    return (
+        <div>
+            You have used {seconds} seconds
+        </div>
+    );
+}
```
kiko-bkr, line 17: Good job!

kiko-bkr (changes requested):
The code you have written is correct with the exception of one small mistake. If you fix that, the rest of the code will work as you expected. 😄 Plus there is another small correction in the checkbox callbacks, refer to the comments for details.

PR #271, opened 2024-03-24
react1-week2/Ihor

IhorKulytskyi:
Homework week2


SvitlanaBurlaka (commented):
Good job! ⭐ I like that you implemented the delete feature! Just one important note: when the application is small, you might think that there is no need to move functions like 'addTodo' and 'toggleDone', etc., outside of App.jsx file. However, it would be good practice to start keeping your App.jsx file as clean as possible from now on. Here's why:

1. You might want to reuse functions in other components, making your code reusable.
2. Each file or component should have a single responsibility.
3. It will make your code more readable. This is crucial when you work in a team, and another person would need to go and fix something. (It's good to fix just one broken component, instead of touching the main App.jsx file where there can be a lot of important things, right?)

PR #274, opened 2024-03-28
react1-week3/niloufar

Niloufar97:


7_react/react1/week3/TodoList/src/App.tsx
```diff
@@ -0,0 +1,57 @@
...
+    return data;
+  }catch(error){
+    console.log(error)
+  }
+}
+function App() {
+  const {data : initialTodo, error} = useSWR(url , fetcher);
+  
+  const [todos, setTodos] = useState<Todo[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
```
aina21, line 30: You can utilize the isLoading status directly from useSWR instead of creating a separate state for isLoading.

7_react/react1/week3/TodoList/src/App.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  useEffect(()=>{
+    if (initialTodo) {
+      setTodos(initialTodo);
+      setIsLoading(false);
+    }
+  }, [initialTodo]);
+
+  const deleteTodo = (id: number) => {
```
aina21, line 39: I suggest moving the deleteTodo function into the TodoList component as you're not depending on it out of the component scope.

7_react/react1/week3/TodoList/src/components/TodoForm.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  const changeTodoInput = (e: ChangeEvent<HTMLInputElement>) => {
+    setNewTodoItem({ ...newTodoItem, description: e.target.value });
+  };
+  const changeDeadLineInput = (e: ChangeEvent<HTMLInputElement>) => {
+    setNewTodoItem({ ...newTodoItem, deadline: e.target.value });
+  };
+
+  const addNewTodo = () => {
```
aina21, line 32: You could add a condition to check if the new item is empty before adding it to the list. This prevents empty items from being included.

aina21 (approved):

PR #275, opened 2024-03-29
react1-week3/anastasiia

humenyuk16:
React week3 homework:

Add a deadline to the todo item
Create new item using description and deadline that the user inputs
Possibility to update a todoitem
Create proptypes for at least one of the components
Create a border component that wraps a component in a black border


7_react/react1/week3/vite-project/src/components/AddTodo.jsx
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from 'react'
+
+function AddTodo({addTodo}){
```
NVedsted: Great idea to extract this into its own component with its own state! 👍 Perhaps the validation logic should reside in here as well? I lost my changes when my input failed validation and had to type it again. 🥺 

You can also leverage the browser to do some of the validation! If you put the `required` attribute onto `<input>`s, the browser will tell the user to fill the fields out before submitting the form. 😄

7_react/react1/week3/vite-project/src/components/Timer.jsx
```diff
@@ -0,0 +1,30 @@
+import React, { useState, useEffect } from 'react';
+
+const Timer = () => {
```
NVedsted, line 3: Nice work! ⭐ It does seem like you do not need the `startTime` state as it is only listed as a dependency which is not in use in your effect. A tiny nitpick is that the `React` import is unused. 🙂 Love the use of a helper function to format the time to keep the HTML clean! 🤩

7_react/react1/week3/vite-project/src/components/TodoList.jsx
```diff
@@ -0,0 +1,65 @@
+import PropTypes from "prop-types";
+import Border from "./Border";
+
+function TodoList({ todos, toggleDone, deleteTodo, editTodo }) {
```
NVedsted, line 4: This component is quite crowded, but it gets the job done. 💪 Can you think of a way to split the functionality up a bit to make it more digestible? 😄

7_react/react1/week3/vite-project/src/App.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+const App = () => {
+  const [todos, setTodos] = useState([]);
+  useEffect(() => {
+    fetch(
+      "https://gist.githubusercontent.com/benna100/391eee7a119b50bd2c5960ab51622532/raw"
+    )
+      .then((response) => response.json())
+      .then((data) => {
+        setTodos(data);
```
NVedsted: If this `fetch` was very slow, it could potentially overwrite todos created by the user! 😮 Can you think of a way to avoid this happening?

7_react/react1/week3/vite-project/src/App.jsx
```diff
@@ -0,0 +1,85 @@
...
+    <div className="app">
+      <Timer />
+      <h1 className="app_title">What I need ToDo? &#129300;</h1>
+      <p className="app_description">Create your own list</p>
+
+      {todos.length === 0 ? (
+        <p>No items</p>
+      ) : (
+        <TodoList
+          todos={todos}
```
NVedsted, line 85: The prop types for `<TodoList>` requires each todo to have a boolean containing whether it is done. This is not the case for the todos provided by the `fetch`. Perhaps some processing should be done to those fetched todos or the prop types should be modified. What do you think is a good solution? 😄 

This also causes some issues with your checkboxes as React assumes them to be uncontrolled when their value is undefined, but then they get a value when you check them off. [You can read more about this issue here.](https://react.dev/reference/react-dom/components/input#im-getting-an-error-a-component-is-changing-an-uncontrolled-input-to-be-controlled)

NVedsted (changes requested):
Very nice work! 👍 Really liked your styling. 😄 There are a few issues and some errors in the browser console when visiting the page, but overall a very well functioning app. 🎉 

Feel free to reach out if you have any questions or issues. 🙂

7_react/react1/week3/vite-project/src/components/TodoList.jsx
humenyuk16: @NVedsted I do not know actually how to do it correct, or you mean separate it into smaller components in the separate files? Like TodoItem, TodoEdit,TodoList smth like that? :)

7_react/react1/week3/vite-project/src/App.jsx
humenyuk16: @NVedsted  I did some changes,
 you can have a look?🥺

7_react/react1/week3/vite-project/src/components/TodoList.jsx
NVedsted: Yes, exactly like that. 😄

7_react/react1/week3/vite-project/src/App.jsx
NVedsted: Nice with some additional error handling and an easier to read async function! I think you are on the right track with the `loading` state, but you are not using the value anywhere at the moment. If you only show the todo list once data has loaded, then we get around the issue. 😄

NVedsted (approved):
I think you got around to address most of my comments; good work! Very fast as well. 😮 There is the one thing with the `loading` state, but I wouldn't spend too much time on that. 😄

PR #277, opened 2024-03-30
react1-week1/sanaz

Sanazre:


7_react/react1/week1/vite-project/src/components/ItemListDisplay..jsx
```diff
@@ -0,0 +1,9 @@
+const ListingItems = (props) => {
```
SvitlanaBurlaka, line 1: I would suggest to have the same name for component and file where component is located, to avoid confusion. For example it can be `const = ItemListdisplay` the same way as file named.

7_react/react1/week1/vite-project/src/components/TodoListRenderer.jsx
```diff
@@ -0,0 +1,18 @@
+import ListingItems from "./ItemListDisplay.";
+
+function IteratingTodoItemsArray({ toDoItems }) {
```
SvitlanaBurlaka, line 3: here is the same, change the name for component and your file, so it is matching and descriptive.

SvitlanaBurlaka (commented):
Good job! ⭐  in description for homework you need to have more than two components, so add one more component to your project. HINT: it can be e.g. Header component where you can have just 'TODO List' text ! 😻

PR #278, opened 2024-03-30
react1-week2/Helena

helena69hme:


SvitlanaBurlaka (commented):
Hi Helena! 
Folder where you can have your components better to name it as 'Components' not 'ingredients'. Creativity is good, but it is a standart practice to name it as 'Components'. 
I am also wondering, how should I run the project on my local machine? to be able to run it locally I need package.json file which is not here.

PR #279, opened 2024-03-30
react1-week3/katayoun

katayouny:


7_react/react1/week3/vite-project/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "dev": "vite",
+    "build": "vite build",
+    "lint": "eslint . --ext js,jsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "prop-types": "^15.8.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "unfetch": "^5.0.0"
```
NVedsted: `unfetch` is unused and could be removed. If you've looked into it, can you explain what this package would help you with and how it could have been used? 😄

7_react/react1/week3/vite-project/src/components/addNewToDo.jsx
```diff
@@ -0,0 +1,168 @@
+import { useState, useEffect } from "react";
+import ListItemWithBorder from "./listItemWithBorder.jsx";
+
+const API_URL="https://gist.githubusercontent.com/benna100/391eee7a119b50bd2c5960ab51622532/raw";
+
+function AddNewToDo() {
```
NVedsted, line 6: The name of this component is perhaps a bit misleading as its functionality goes quite beyond just adding new todos. 😄 It is quite a mouthful to take in! Can you think of a way to simplify the component by splitting it up into multiple components? 🙂

7_react/react1/week3/vite-project/src/components/addNewToDo.jsx
```diff
@@ -0,0 +1,168 @@
...
+    );
+    setTodos(updatedTodos);
+  };
+
+  const handleDeleteTodo = (id) => {
+    setTodos(todos.filter((todo) => todo.id !== id));
+  };
+
+  const handleEditClick = (id, description) => {
+    setEditModeId(id);
```
NVedsted, line 55: This choice of state and component layout has some effects that may or may not be desirable. The most obvious is that you can only edit one todo at a time; that might be fine, but could also be unnecessarily restricting. As alluded to further up, it might be a lot easier to do if you split the component up a bit. If you would like to support this, can you think of how to do it? 🙂

I did discover a bug: if you edit a todo and delete it before pressing update, you can no longer edit any todos. Can you see why? And can you find a solution that both fixes this issue and perhaps makes it easier to support multi-editing? 😄

NVedsted (changes requested):
Awesome work; a neat app with a pleasant look! 🌟 Generally, I'd suggest sticking to the convention of naming your component files exactly as the component similar to `<App>` in `App.jsx`, e.g. `<Timer>` in `Timer.jsx`. 😄 

I have left a few comments; feel free to reach out if you have questions or encounter issues. 🙂

7_react/react1/week3/vite-project/package.json
katayouny: Hi
Thank you so much for your review. I was trying different methods for fetching, and at last, I chose the current one I have with .then. I couldn't figure out how it works and I got some errors 😅

7_react/react1/week3/vite-project/src/components/addNewToDo.jsx
katayouny: I will change it to ToDoManaging or I will make separate components :)

7_react/react1/week3/vite-project/src/components/addNewToDo.jsx
katayouny: Regarding this part: 'If you would like to support this, can you think of how to do it? ', could you please give me a hint, thank you 😊

'And can you find a solution that both fixes this issue and perhaps makes it easier to support multi-editing?' I will try that 😊

katayouny:
Hey @NVedsted 
I have made some changes. 
Still have not created separated components in separate files :).
Thank you.

PR #282, opened 2024-04-02
react1-week2/gizem

GizemSavci:
https://github.com/HackYourFuture-CPH/class27-homework/assets/129002582/c9eac9a8-ab80-4b7a-a352-5c81af9a5516


7_react/react1/week2/todo-app/src/components/Timer/Timer.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+
+const Timer = () => {
+
+    const [seconds, setSeconds] = useState(0);
+    const [minutes, setMinutes] = useState(0);
+    
+    useEffect(()=>{
+        const timer = setInterval(()=>{
+            setSeconds(prevSeconds => {
```
NVedsted: Awesome work with the split into minutes and seconds! 🎉 Much more appealing to look at.

There is a small issue with the updater function for seconds: it is not pure! It has a side-effect since it calls `setMinutes`. You might have seen when running in development mode, that your minute counter increments twice every time 60 seconds pass. This is due to it running it strict mode where it runs functions supposed to be pure twice to provoke purity issues during development. [You can read more here.](https://react.dev/reference/react/useState#my-initializer-or-updater-function-runs-twice)

Can you think of a way to do this with only one simple state and not two? Then it becomes a lot easier to manager. 😄

7_react/react1/week2/todo-app/src/components/TodoItem/TodoItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+        onChange={handleChecboxChange}
+      />
+      <Task task={todo.task} isChecked={isChecked}/>
+      <Deadline deadline={todo.deadline} />
+      <button className="remove-button" onClick={handleRemoveClick}>x</button>
+    </div>
+  );
+}
+
+TodoItem.propTypes = {
```
NVedsted, line 32: Very nice to set some types and the like so we can catch pesky type errors and missing props! 👍 `Deadline`, `Task`, and `TodoList` are missing them though. They are not required, but consistency is key! 😄

7_react/react1/week2/todo-app/src/components/TodoItem/TodoItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+import './TodoItem.css';
+function TodoItem({ todo, removeTodo }) {
+  const [isChecked, setIsChecked] = useState(false);
+
+  const handleChecboxChange = () => {
+    setIsChecked(!isChecked);
+  }
+
+  const handleRemoveClick = () => {
+    removeTodo(todo.id);
```
NVedsted, line 14: Good use of callback patterns! ⭐ I would suggest going with the convention of `onSomeAction` to make it clear this is akin to listening to an event in the component.

NVedsted (changes requested):
Great work! 🌟 Was a joy to setup your app and see it in action. 😄 Feel free to reach out if you have any questions. 🙂

7_react/react1/week2/todo-app/src/components/Timer/Timer.jsx
GizemSavci: I created one state `elapsedTime`. Can you check from my commits:`Fix Timer bug`? Thank you for helping with this bug🙏

7_react/react1/week2/todo-app/src/components/TodoItem/TodoItem.jsx
GizemSavci: Thank you🙏 I just used PropTypes for `Deadline`, `Task` and `TodoList`.

7_react/react1/week2/todo-app/src/components/TodoItem/TodoItem.jsx
GizemSavci: but React documentation has examples with `handleStuff`. Sorry, I couldn't understand what to do🙈

PR #286, opened 2024-04-07
react1-week3/GizemSavci

GizemSavci:


7_react/react1/week3/todo-app/src/App.jsx
```diff
@@ -0,0 +1,41 @@
...
+import { useEffect, useState } from "react";
+import useSWR from "swr";
+import { TodoForm } from "./components/TodoForm/TodoForm";
+
+const fetcher = (...args) => fetch(...args).then((response) => response.json());
+
+function App() {
+  const [todos, setTodos] = useState([]);
+
+  const { data, error } = useSWR("https://gist.githubusercontent.com/benna100/391eee7a119b50bd2c5960ab51622532/raw", fetcher);
```
BrantalikP, line 12: well, this is cool. good job!  Here is some really popular alternative if you are interested :) https://tanstack.com/query/latest

7_react/react1/week3/todo-app/src/App.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+  useEffect(() => {
+    if (data) {
+      setTodos(data);
+    }
+  }, [data]);
+
+  if (!data) return <div>Loading...</div>;
```
BrantalikP, line 20: we should show the `loading` state only if the data is really loading and it is empty, otherwise, it could confuse the user by letting him stuck on the loading page even if there is no data. please add `isLoading` prop from `useSwr` as well :)

7_react/react1/week3/todo-app/src/App.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+  if (error) return <div>Error: {error}</div>;
+
+  const addTodo = (todo) => {
+    setTodos((prevTodos) => [...prevTodos, todo]);
+  }
+
+  const removeTodo = (todoIdToRemove) => {
+    setTodos(todos.filter(todo => todo.id !== todoIdToRemove));
```
BrantalikP, line 29: to keep the code consistent, please use the same approach as you did above in addTodo with `prevTodos`

7_react/react1/week3/todo-app/src/components/BorderWrapper/BorderWrapper.jsx
```diff
@@ -0,0 +1,19 @@
+import { Children } from 'react';
+
+function BorderWrapper({ children }){
+    return (
+        <div style={{
```
BrantalikP, line 5: there is no need for inline styling, please use CSS styles from CSS file

7_react/react1/week3/todo-app/src/components/DatePicker/DatePicker.jsx
```diff
@@ -0,0 +1,21 @@
+import { useState } from 'react';
+import DatePicker from 'react-datepicker';
+import 'react-datepicker/dist/react-datepicker.css';
+
+function DateSelector({ selectedDate, onDateChange }) {
+  return (
+    <div className='App'>
```
BrantalikP, line 7: I've noticed that you are using this class in multiple places but I cannot see such a style anywhere if it's not doing anything please remove it :)

7_react/react1/week3/todo-app/src/components/Deadline/Deadline.jsx
```diff
@@ -0,0 +1,15 @@
...
+    return (
+        <div className="deadline">
+            <p>Deadline: {props.deadline}</p>
+        </div>
+    );
+}
+
+export default Deadline;
+
+Deadline.propTypes = {
```
BrantalikP, line 13: well done for using Proptypes 👏

7_react/react1/week3/todo-app/src/components/TodoForm/utilities.js
```diff
@@ -0,0 +1,6 @@
+let nextId = 0;
+export function generateId() {
```
BrantalikP, line 2: This approach is not safe, it would work only till the user won't refresh the page. 

Lets say that you would store the user's data, 
1. The user comes to the page and will create a todo "some todo" id: 0 (nextId current value)
2. He decided that he will the rest later so he closes the browser, then he comes back after some time and create another todo, but because the previous instance of the page was terminated the `nextId` start from 0 again

There are other ways how to achieve a uniq ID: 
1. using timestamp, you can use Date API to get the current timestamp, to make it more robust, you can combine it with the todo description
2. Math.random method
3. maybe overkill for the app of this scale, but the safest approach is using uuid library https://github.com/uuidjs/uuid

7_react/react1/week3/todo-app/src/components/TodoForm/TodoForm.jsx
```diff
@@ -0,0 +1,46 @@
...
+        setSelectedDate(date);
+    }
+
+    const handleSubmit = (event) => {
+        event.preventDefault();
+        if (description.trim() && selectedDate) {
+            const todo = {
+                id:generateId(),
+                description: description,
+                deadline: selectedDate.toISOString().split('T')[0],
```
BrantalikP, line 26: This will also cause some issues with in your timezone as well. If you've noticed, the displayed date will differ from the selected one by one day. This is caused by the `toISOString` method, which converts the date to UTC time. Alternatively, you can use this approach if you know that only people from Denmark will use it.

```
selectedDate.toLocaleDateString("en-CA", {
        timeZone: "Europe/Copenhagen",
      });

```

BrantalikP (changes requested):
Besides few things, as always @GizemSavci amazing job!I can see all the hard work behind it and I appreciate that you are not afraid to do that one extra step to achieve better results. Big Kudos 👏

PR #294, opened 2024-04-12
react2-week1/katayoun

katayouny:


7_react/react2/week1/vite-project/src/components/UserSearchBar.jsx
```diff
@@ -0,0 +1,38 @@
...
+      <h4>User Search</h4>
+
+      <div>
+        <form>
+          <label>
+            Search for user name:
+            <input
+              type="text"
+              placeholder="insert user"
+              style={{
```
aina21, line 18: Try to avoid using inline styles when possible. Instead, consider moving the styles to a CSS file.

7_react/react2/week1/vite-project/src/components/UserSearchResult.jsx
```diff
@@ -0,0 +1,23 @@
+import { useContext } from "react";
+import { userContext } from "./UserSection";
+
+function UserSearchResult() {
+  const { query, loading, error, searchResult } = useContext(userContext);
+
+  return (
```
aina21, line 7: Consider simplifying the conditional logic. You could use a helper function or separate the outputs using if conditions, like so:

```jsx
if (loading) return <div className="loading">Loading...</div>;
```

aina21 (approved):

7_react/react2/week1/vite-project/src/components/UserSearchBar.jsx
katayouny: Hey @aina21 
Yes, thank you.

katayouny: @aina21 
Could you please tell me why I can't see the search results? I get no error. I tried to see what I would get in fetch using console.log(data), and then I could see the items. Therefore I used data. items. However, I would get no search results in the browser. Could you please help me where the problem is? Thank you so much.

PR #297, opened 2024-04-14
react2-week1/maryam

MaryamAarhus:


7_react/react2/week1/Github-user-Search/src/GithubUsersContext.jsx
```diff
@@ -0,0 +1,34 @@
...
+        setLoading(true);
+        setError('');
+        try {
+            const response = await axios.get(`https://api.github.com/search/users?q=${query}`);
+            setUsers(response.data.items);
+            if (response.data.items.length === 0) {
+                setError('No results...');
+            }
+        } catch (err) {
+            setError(`Error fetching: ${err.message}`);
```
aina21, line 24: You should set loading to false when you catch an error.

7_react/react2/week1/Github-user-Search/src/GithubUsersContext.jsx
```diff
@@ -0,0 +1,34 @@
...
+        }
+        setLoading(false);
```
aina21, line 26: Moved setLoading(false) to a finally block to ensure it’s executed regardless of whether the try block succeeds or the catch block is executed.

aina21 (approved):

PR #301, opened 2024-04-20
react2-week1/olesia

AlesyaMazurenko:


BrantalikP (changes requested):
Hey @AlesyaMazurenko It looks like you pushed an incorrect/incompleted branch. I cannot see the code related to the react2/week1 assignment. Please check it one more time and remove all unrelated code (meaning also the code for the other weeks) Thank you!

